PR #17, opened 2015-01-20
Fixed ImportError in `admin_tree` occurring in current Django `master` 

kezabelle:
because `django.contrib.admin.util` became `django.contrib.admin.utils`


nikolas:
Looks good -- this should be merged.

g3rd:
Also, looks good to me. Any ides when this will be merged?

PR #18, opened 2015-05-18
Allow natural primary keys - see #67

rhunwicks:
As discussed in https://bitbucket.org/tabo/django-treebeard/issue/67/treebeard-admin-cannot-work-with-custom-pk I think it is now safe to use `self.instance._state.adding` to determine whether the object already exists in the database


rhunwicks:
Is there anything additional that I need to do in order for you to merge this?

Thanks

Roger

PR #19, opened 2015-05-18
Added support for SQL Server

rfosterslo:
Added SQL Server-specific syntax for update query.


mkoistinen:
I just thought I'd pop in here to say that django CMS would also benefit from Treebeard's support for MS SQL Server. Also, considering recent announcements for MSSQL on Linux, this will become increasingly in demand.

tabo:
I could merge this, but I have no way to know if it's actually working.

jrief:
Is there a way to add an MSSQL server to the testing matrix on Travis-CI?

mvantellingen:
This should be supported now via e228fe87d38e3ad58fe42dae2bf801522037c60f

PR #20, opened 2015-05-25
Allow subclassed models to be created using MoveNodeForm - see #80

rhunwicks:


rhunwicks:
Is there anything additional that I need to do in order for you to merge this?

Thanks

Roger

PR #30, opened 2016-06-14
Allow natural primary keys

kolanos:


kolanos:
@tabo Esto es en referencia a #18, por favor hágamelo saber si hay problemas .

johanneswilm:
Hey @jrief This PR is now 4.5 years old and we still urgently need this in an official release. Would it be possible to take a look at it and create a release with it some time in the not so distant future?

johanneswilm:
ping @tabo @kolanos We need a release with this ASAP and it has been sitting here for 4.5 years now.

johanneswilm:
Hey @solarissmoke do you have write access rights for this repository? I see you are the last ones committing things here.

johanneswilm:
Hey @mvantellingen and @yakky - it looks like you have access here and you are still active on github, whereas @tabo seems to be gone. Is there a plan for how django-treebeard is to continue? Or should we just fork and release an alternative package?

solarissmoke:
@johanneswilm I have commit access to the repo, but no access to PyPi, which is why I have not merged #186. I've reached out to the owner numerous times to try and get this access, without success. At this point I think the best option is for someone to fork the package. I'm happy to help with maintenance.

johanneswilm:
@solarissmoke I see that Pypi now apparently has implemented [PEP541](https://www.python.org/dev/peps/pep-0541/), so you could also try to claim the `django-treebeard` name there by filing a ticket here: https://github.com/pypa/pypi-support/issues/new?assignees=&labels=PEP+541&template=pep541-request.md&title=PEP+541+Request%3A+PROJECT_NAME

But if you tried that already, or we will have more problems in that we cannot add more collaborators, we may be better of trying to create a new project. Any suggestion for a name? Can you create new repos under the django-treebeard organization?

tabo:
Hi all, just catching up on all this. We have 4 other maintainers in pypi, have they become inactive?

johanneswilm:
Hey @tabo - it looks like @jrief has done some things over the past year. But does he have access to the Pypi repo? And I don't know how much time he has.

tabo:
@johanneswilm yes, @jrief is a maintainer in pypi. I just saw in #178 that he is planning a release today.

johanneswilm:
@tabo @jrief - if you don't have time to continue maintaining this, could you let @solarissmoke on as co-maintainer?

johanneswilm:
@tabo - yes, but that was 14 hours ago. And for example this PR has been sitting here now since June 2016.

johanneswilm:
I see now that there was a 4.4 release to Pypi today. That's great, but unfortunately there are still lots of PRs that haven't been touched for a long time, like this one. We have been running a forked version of django-treebeard with about 500 of our own commits (for python/django upgrades, etc.) of the main version because this hasn't been merged for the past 4.5 years. It has become a complete nightmare to try to run our own version. We are really desperate to get back to a version that is on pypi.

It would be really helpful if you could take on someone else, like @solarissmoke, who is willing to spend the time maintaining this.

jrief:
@johanneswilm unfortunately the issue describing the problem which this pull request shall solve, is gone (hosted on Bitbucket) at the time.
This library is a dependency of so many important Django projects, that my main focus is stability rather than new features – specially those I don't understand. I therefore really would like understand why a new features must be added before merging arbitrary pull request.

johanneswilm:
Hey @jrief - ok. Let me try to do that briefly - django-treebeard works well for when the pk is an automatically assigned number. It doesn't work so well for other primary keys. So there is another PR for UUIDs as pks [1]. We use chars. In that other PR [1] you asked the users to please ping another maintainer - but it seems like you are the only one around. 

Are you suggesting it would be better to just fork this repository instead and create a new package? Why not make a version 5 with the new functionality and let every old package that doesn't want anything changed depend on this version 4.x?

[1] https://github.com/django-treebeard/django-treebeard/pull/93/files

tabo:
Thank you for the comments. I am glad adding @solarissmoke to pypi if he wants to contribute, but note that I have to agree with @jrief here: there are projects depending on the stability of this library, so caution merging changes is important.

I am not up to date with changes in Django so I can't really tell how solid this change is, but I see it is accessing a _state private property. Is django providing a better API to accomplish what we need in the years after this PR was made? If not, is this private property widely used in other Django projects as to know if django core devs will be cautious breaking it? Also: are the tests comprehensive enough?

I like @johanneswilm 's idea of starting a new 5 branch for breaking changes, that's why we are using semver, but this PR doesn't seem to be the case for it. Mostly, seems to be a case of the community reaching to maintainers for a solution. Since @solarissmoke has community trust by having git write access, I'd be happy to give him pypi access.

Cheers!

rhunwicks:
@jrief @tabo 

As @johanneswilm said, this code is needed when using a Django model with a natural primary key, e.g. a character string rather than an auto-incrementing integer.

For example, we maintain a Django model that contains a list of units of measure, such as Kilogram, Liter, etc. Those things have natural abbreviations, and so it makes sense to use those abbreviations (kg, L, etc.) as the primary key. That way, when we have another model with a ForeignKey to UnitOfMeasure, the value in the other table will be, say, `kg` rather than an integer such as `35`. With an auto-incrementing primary key, you always need to join the other table in SQL in order to fetch sufficient information to make sense of the value. But with a natural key, we can look at the `unit` attribute in the model instance and see that it is `kg` and avoid the join. It obviously isn't appropriate for most tables, but for those models that are reference metadata with low row counts and obvious identifiers it is very helpful. We use it for things like Country (DE, FR, GB, etc.) and Currency (EUR, GBP, USD, etc.).

```python
class Country(Model):
    """
    A Country (or dependent territory or special area of geographical interest) included in ISO 3166.
    """
    iso3166a2 = models.CharField(max_length=2, primary_key=True, verbose_name="ISO 3166-1 Alpha-2")
```

This setup is well-supported by Django.

We also have some models that need to be both hierarchical and use natural primary keys. For example, we maintain a list of products that uses the UN's CPC v2 (https://unstats.un.org/unsd/classifications/Econ/CPC.cshtml) codes.

```python
class ClassifiedProduct(MP_Node, Model):
    """
    A product such as a commodity or service classified using UN CPC v2 codes.

    See http://unstats.un.org/unsd/cr/registry/cpc-2.asp for more information

    """

    cpcv2 = models.CharField(max_length=8, primary_key=True, verbose_name="CPC V2")
```

Unfortunately, this approach doesn't work with Treebeard without this MR, because in a model with a natural primary key, you provide the key when you first save the model in the database, and that makes Treebeard raise `NodeAlreadySaved`.

The approach used in this MR also checks the `_state` variable to avoid the exception if we are sure we are adding a new record.

We have been running this code since 2015, and had actually submitted it as an MR when development was still on BitBucket (https://bitbucket.org/tabo/django-treebeard/issues/67/treebeard-admin-cannot-work-with-custom-pk if anyone has access to an archive) - see https://github.com/rhunwicks/django-treebeard/commit/149c35c245a4fee73f4609ce02135932f6b937db. Obviously, I can't guarantee stability, but we have been running it in production for a really long time! Also, the code is not very invasive, and only adds an additional check before raising the exception in a small number of places. I think that the fact that the MR has been open since 2016 and would still merge cleanly also suggests that it isn't invasive.

I appreciate the concern about using the private `_state` variable to make the check, but I don't think it is a private variable any more. It used to be private, and hence has the underscore prefix, but it is documented in modern Django, see https://docs.djangoproject.com/en/3.1/ref/models/instances/#state

I am happy to answer any follow on questions about the history of this change.

rhunwicks:
The wayback machine to the rescue: https://web.archive.org/web/20200723070912/https://bitbucket.org/tabo/django-treebeard/issues/67/treebeard-admin-cannot-work-with-custom-pk

solarissmoke:
Hi @tabo!

> Since @solarissmoke has community trust by having git write access, I'd be happy to give him pypi access.

Please do! My pypi username is also solarissmoke. I'd also be grateful if you can add me as a member of this organisation (I'm currently just a collaborator on this repo).

As regards the question of stability of the library, my view is that this can largely be addressed by three things:

1. Thorough unit testing, which this package has done an excellent job of so far.

2. Careful review of changes before they are merged, which I'm happy to spend a bit of time on.

3. Proper use of semantic versioning so that changes that *are* potentially going to break things are shipped as a major version, with appropriate development releases shipped early, so that projects are given a strong signal that they need to test before upgrading. 

In my view, if this package is to continue being a valuable part of the Django ecosystem then it needs to do more than just maintain existing functionality, and if people like are willing to contribute useful new features then this should be encouraged.

jrief:
@johanneswilm 
>  it makes sense to use those abbreviations (kg, L, etc.) as the primary key

No, absolutely not! You should never abuse your values as primary keys, nor should you use strings for this purpose. Primary and Foreign Keys are just addresses and pointer, just as you would use them in languages such as C, etc. They are abstract entities, which should only be used for the purpose of addressing stuff.

I worked on too many projects where someone believed that by using a column with a unique value, she/he can save some bytes by using that values as a primary key. It always turned out to be a bad idea, so please don't do it.

I'm therefore very reluctant to merge such a hack.

johanneswilm:
Hey @jrief, the person you are quoting is @rhunwicks, not me. Also he has decades of experience working on programming projects and for whatever reason has come to different conclusions.

I will not judge here what is the best practice for primary keys. Personally, I tend to just use the built-in primary key system like also you suggest. But Django is also supporting these other PKs. It's not a hack, but rather an officially supported feature of Django, so there will always be projects out there that make use of that feature. If you think users should be prohibited from doing so, I would argue that the place where this should be done would be the central django repository, not by not accepting a PR on a dependent project like this one.

rhunwicks:
@jrief there are many arguments both for and against the use of natural primary keys versus surrogate keys.

The main argument against natural keys if that if you ever have to change the keys then the cost and complexity of doing so is enormous. Hence the guidance to use surrogate keys in most instances. But identifiers issued by international bodies like ISO or the UN Statistics agency are effectively permanent. Once they have issued an identifier, if it no longer becomes required they add it to a list of retired codes, so it will never be issued again. I cannot imagine a circumstance under which we will not be able to use FR and DE as the identifiers for France and Germany.

The main argument in favor of natural keys is that of meaningful representation in related tables, and the reduced the number of joins in complex queries. Those issues aren't easily visible in most systems written in Django, with relatively low row counts, and the ORM automatically joining the tables and showing a meaningful representation of the ForeignKey field. But those issues still do exist for users who are not accessing the database through Django. In our case, the main table contains 16 millions rows of data, and we need to run complex analytic SQL queries on (window functions, etc.) and so unnecessary joins do have a noticeable performance impact.

There are not many situations where a natural key is preferred over a surrogate key, but there are a few, and I don't agree that we should refuse to use them in those situations because other developers have used them inappropriately in other situations.

As @johanneswilm says, manually-specified natural keys is entirely supported by Django, and I think that Treebeard should support the full range of Django models where possible.

I think that as it stands, Treebeard would also struggle with a system where the data was imported from another system (where it might have a surrogate primary key, but one that is not set by Django). Django does support insert rows into a database that uses an auto-incrementing `id` column, but specifying the `id` column on insert. That is also a rare situation but sometimes necessary when migrating applications.

Treebeard might also struggle with applications that use a `UUIDField` as a primary key. That's also currently unusual but is gaining traction. Postgresql handles them very efficiently.

johanneswilm:
> I cannot imagine a circumstance under which we will not be able to use FR and DE as the identifiers for France and Germany.

Germany splitting into 16 smaller republics that are at war with one-another and half of France joining the UK as a member state. :) 

But yes, agreed, the likelihood of that is too small to take it consideration and there can be other issues, such as performance, that lead to different decisions concerning PKs. @jrief  Note that @rhunwicks has a background from working with databases [1] so it's not like he is a noob unaware of the inner workings of databases. But at times it does mean that he prioritizes other issues---due to possible database performance concerns---than what I as someone with a pure Django/Python background would do by default.

[1] https://www.linkedin.com/in/roger-hunwicks-0968a365/?originalSubdomain=bd

rhunwicks:
This also seems related to https://github.com/django-treebeard/django-treebeard/pull/182/files which is also using `_state.adding` to avoid the Exception when creating model instances, because of a UUID primary key, I think. That PR also fixes the problem of the MoveNodeForm coercing the pk to an integer, which we have been handling using a custom form.

https://github.com/django-treebeard/django-treebeard/pull/93 from 2018 also does almost exactly the same thing to enable UUID fields.

I.e. 3 of the 11 open PRs are requesting support for user-defined primary keys.

rhunwicks:
#93 was approved by auvipy in December 2019, but not merged. If #93 was merged it would make both this MR and #182 redundant, I think.

tabo:
> I worked on too many projects where someone believed that by using a column with a unique value, she/he can save some bytes by using that values as a primary key. It always turned out to be a bad idea, so please don't do it.
> 
> I'm therefore very reluctant to merge such a hack.

Many lifetimes ago, when Django was still a new promising framework, I worked for a company heavily involved in Postgres, and they hired a couple of Postgres core developers. They asked us to pick their brains, so I did. When I asked one of them about his thoughts on Django's ORM, he told me that he was ok with the idea of ORMs, but he was very against its design of surrogate keys (I'm talking about really old Django here). He has been writing about this topic for years on his blog, I think his most recent article is this one: http://www.databasesoup.com/2015/03/primary-keyvil-reprised.html - His conclusion is:

> As I said earlier, it’s not using autonumber surrogate keys (which are a necessary, pragmatic evil) but misusing them that causes pain, late nights and budget overruns. Just make sure that you’re using integer keys in the right ways and for the right reasons, and always ask yourself if an autonumber surrogate key is really needed in each table where you put one.

Additionally, Django has supported natural keys for years. Is there something in the implementation that would break Treebeard? I remember being very careful when I wrote initially to not mess with PKs or anything in the user's data. Are there any concerns that we can reproduce with tests? And if there are, can we fix them so we can make these tests pass? I mean, looking at the changes in this PR, all those ifs looking for an existing .pk are just trying to check if the object has been persisted before. 2020 me would have probably extracted all of those into a .is_persisted() or something like that for clarity (in fact, we should do that, with our without this patch).

And @rhunwicks if there are multiple PRs solving the same issue, then there is definitively interest in this problem, but now we come to the issue of either picking the best one, and/or using them as a base for a better solution. I'm mostly concerned with test cases here.

I'm also open to the idea of not merging anything it this will introduce breaking changes.

johanneswilm:
> I'm also open to the idea of not merging anything if this will introduce breaking changes.

If this is a feature that there is a need for but it could potentially introduce breaking changes, then I'd argue the best solution would be to go with a new major version in order to be in line with semver.

However, I don't think that is the case here at all. I don't see how this PR can break anything for anyone using the default auto-incrementing PKs. Also, there is an added test in this PR. What other thing were you thinking of testing, @tabo?

> if there are multiple PRs solving the same issue, then there is definitively interest in this problem, but now we come to the issue of either picking the best one

They actually look very similar. It may not matter which one you choose. All they do, really, is modify the way it checked for whether `_state.adding` is `True`. #182 removed the check for `.pk` entirely, and I think that may actually be the best solution as `_state.adding` should always be `True` if there is no `.pk` value in the auto-incremeanting model. This may also be more future proof, as Django 3.2 allows custom auto-incrementing PK functions [1] and not all of them may give the `.pk` a falsy value by default.

[1] https://docs.djangoproject.com/en/dev/releases/3.2/#customizing-type-of-auto-created-primary-keys

rhunwicks:
I agree with @johanneswilm that I don't think this will break anything. That said, all of the 3 MRs have different advantages:

* #30 (this one) fixes natural keys for AL, MP and NS trees and includes a test for a model with primary key already set, but it doesn't alter the forms to prevent them coercing the key to `int`.
* #93 only fixes MP trees and leaves AL and NS trees alone, but it includes a specific `MP_TestNodeUuid` model in the tests so it probably has better test coverage than this one. It removes coercion from the form, which means they work for non-integer keys, but could conceivably break an existing application that is relying on them.
* #182 fixes AL, MP and NS trees and includes a `MP_TestNodeCustomId` test model in `BASE_MODELS`. It fixes the forms, but seemingly by allowing coercion to the actual correct field type, which makes it more less likely to break any existing implementations. It is also the most recent MR, and so we could hope that the author (@obayemi) would be willing to answer questions and/or make changes if requested - it currently assumes the primary key will be called `id` even if it is a `UUIDField`, which won't work for our use case. It also updates `tox.ini` for Python 3.8 and 3.9 and adds a `posargs` to the `pytest` command - I don't know if that bothers anyone? It also removes the pk check entirely - and only relies on `_state.adding`. I think that is fine, and how Django documents that `_state` works, but it is a marginally "bigger" change than the other 2 MRs introduce.

I think that #182 is the best MR, but needs to be updated to use a different name to `id` for the primary key field in `MP_TestNodeCustomId`. I'm happy to try and engage with @obayemi to ask him to make those improvements so that it can replace the other 2 MRs, assuming that the maintainers are willing to merge it when it is ready.

tabo:
Ok guys can we file a separate ticket for this? This discussion is way to broad for a PR that doesn't even have a discussion :)

jrief:
@tabo Thanks for that [article](http://www.databasesoup.com/2015/03/primary-keyvil-reprised.html) about natural vs. surrogate keys. It clarified a lot for me and I will rethink about using natural keys in some of my other projects.

johanneswilm:
Hey,
I was actually just thinking about the main change in #182 . I think it shouldn't be very difficult to combine the three PRs into a single one including all the tests. We don't even need to talk much to the other contributors if they are no longer around. If @rhunwicks hasn't done so yet, I can create a new PR that combines all three and tag the other contributors in case they are still around and want to say something about it as well.

johanneswilm:
Hey, I have now created PR #192 that includes all three PRs and their tests.

Issue #39, opened 2016-11-21
Bring test infrastructure back in shape


yakky:
Maybe using Travis + AppVeyor (for windows testing)?

jrief:
On DuTH, @maikhoepfel declared intention to do this. Is this still the case?

maikhoepfel:
Yes!

maikhoepfel:
A decent test infrastructure is achieved by https://github.com/django-treebeard/django-treebeard/pull/49. I think Windows support can safely be done in a separate PR.

yakky:
\o/

PR #40, opened 2016-11-22
First tweaks

maikhoepfel:
I vowed to get the test suite working with Travis (#39), so I had a little fiddle to get the tests working locally. This is what I achieved so far.


maikhoepfel:
Thanks for the approval. I think this is relatively safe, so I'd go ahead and merge it with one approval? Do you agree, @jrief?

yakky:
I agree.
As a rule of thumb I think that 1 maintainer approval is enough for non code or trivial changes, and require 2 only for larger changes

maikhoepfel:
@yakky I was thinking of saying the same thing, but then my feeling is that that will be a bit of an unachievable luxury in light of the realities of OSS development. ;) We shall see!

PR #42, opened 2016-11-22
Move to Read the Docs

jrief:


yakky:
@django-treebeard/core-maintainers Do we want alabster or standard RTD template?

jrief:
Well, Alabaster is the default, when creating new Sphinx projects using sphinx-quickstart.
And btw. it's beautiful too...

yakky:
The RTD theme is different, though. I'd keep the original one by unsetting the theme in conf.py

PR #48, opened 2016-11-23
Add Travis CI

maikhoepfel:
This PR implements basic Travis support. It tests against the documented versions of Django and Python.

We're not yet testing against Postgres and MySQL, and we're not checking coverage. I'll have a fiddle with that another time. 

I also want to update the tox.ini, but am tempted to drop testing against different database backends there. That would make tox 3 times faster, and I sort of feel right now it's almost useless because it's too slow for me to care. Also, it's a bit of a hassle to get MySQL and PostgreSQL configured appropriately, so I doubt people will care enough.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,29 @@
+# Use the faster container-based infrastructure
+# https://docs.travis-ci.com/user/ci-environment/
+sudo: false
+
+language: python
+
+python:
+    - 2.7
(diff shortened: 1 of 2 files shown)


maikhoepfel:
Ping @yakky @jrief.

maikhoepfel:
Thanks for the quick review!

PR #49, opened 2016-11-23
Test against MySQL and Postgres on Travis

maikhoepfel:
Don't merge please.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,21 +10,34 @@ python:
     - 3.5
 
 env:
-    - DJANGO="Django>=1.7,<1.8"
-    - DJANGO="Django>=1.8,<1.9"
-    - DJANGO="Django>=1.9,<1.10"
-    - DJANGO="Django>=1.10,<1.11"
+    - DJANGO="Django>=1.8,<1.9" DATABASE_ENGINE="sqlite"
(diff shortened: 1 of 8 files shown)


maikhoepfel:
Progress report: Travis is pretty close to working. I'll then rip out tox support for database backends other than sqlite, and package it up into nice commits.

jrief:
Currently the tests fail with Django-1.7 and Python-3.5.
From my point of view we can exclude that test from the matrix, since it is a very uncommon combination.

maikhoepfel:
My friends, thanks for the early review, but this was nowhere near ready! Sorry if I wasn't clear, this was just my scratch work-in-progress PR. I now massaged into something I'm happy to offer for review. Please see the individual commit messages, but generally:

- Django 1.7 support removed
- Travis now also runs against MySQL and Postgres
- Tox works again, but only runs against SQlite (so might actually be useful)
- Coverage still isn't supported, should be added later. 

@jrief: I agree, and having seen that Django 1.7 is not supported any more, decided to drop support for it completely. 

Travis is being a bit slow, so I don't yet know if all tests will pass. I'll keep updating the PR until then, but think it's safe for review, as needed changes will be minor.

maikhoepfel:
Tests pass!

image

jrief:
@maikhoepfel great news, and thanks for taking over that part!

PR #54, opened 2017-02-15
Added helper functions to generate SQL that works on multiple databases

JordanReiter:
django-treebeard, which is currently used by django-cms (among other widely-used applications, I am sure), is not compatible with SQL Server because of the hard-coded SQL commands.

I've wrapped these SQL functions into python functions that generate the correct SQL based on the vendor. I've just written in code for SQL Server engines (assuming that they return a vendor string of "microsoft").

Long term, it would probably be worthwhile to rewrite the code so that it makes use of [Django Database Functions](https://docs.djangoproject.com/en/1.10/ref/models/database-functions/).

### DIFF
--- a/treebeard/mp_tree.py
+++ b/treebeard/mp_tree.py
@@ -17,6 +17,34 @@ from treebeard.exceptions import InvalidMoveToDescendant, PathOverflow,\
     NodeAlreadySaved
 
 
+# The following functions generate vendor-specific SQL functions
+def sql_concat(*args, **kwargs):
+    vendor = kwargs.pop('vendor', None)
+    if vendor == 'mysql':
+        return 'CONCAT({})'.format(', '.join(args))
+    if vendor == 'microsoft':
+        return ' + '.join(args)
+    return '||'.join(args)
+
+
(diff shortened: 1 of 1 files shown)


jrief:
These are great news!

The challenging part is, that we can not run Mircosoft SqlServer on Travis-CI. Does anybody have an idea how to handle this, or should we just skip these tests?

Will you continue to support this database-backend in the future?

mvantellingen:
MSSQL is available via appveyor, we should be able to use that (and add windows testing). See #55

jrief:
Great!

@mvantellingen I had some contact with a service at [cloudpipes.com](cloudpipes.com). However, they have no indefinite free plan for Open Source projects, hence this was not a realy option.

jrief:
@mvantellingen you wrote that you merged this PR. But it's still open. Did I miss something?

mvantellingen:
Oh I meant that I merged a PR I created myself to add appveyor support. The next step is to run the tests against mssql on appveyor

mvantellingen:
Merged

PR #58, opened 2017-03-21
Update test infrastructure

mvantellingen:
This PR updates the way the tests are run on tox, travis and appveyor.

The database engine is now part of the tox environment string, and the database drivers are now installed via tox so that only the required ones are installed. This is required since installing some drivers on appveyor don't work.

Also the mssql engine is added and the changes from #54 so that the mssql tests now pass.

See https://ci.appveyor.com/project/mvantellingen/django-treebeard/build/1.0.41

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -6,10 +6,10 @@
 
 [tox]
 envlist =
-    {py27,py34,py35,py36}-{dj18,dj19,dj110}-{sqlite,postgres,mysql}
+    {py27,py34,py35,py36}-{dj18,dj19,dj110}-{sqlite,postgres,mysql,mssql}
 
 [testenv:docs]
-basepython=python
+basepython = python
 changedir = docs
 deps =
     Sphinx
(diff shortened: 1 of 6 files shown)


mvantellingen:
@JordanReiter can you also take a look at this? :-)

tox.ini
```diff
@@ -19,8 +19,19 @@ commands =
 
 [testenv]
 deps =
-    -rrequirements_test.txt
+    pytest>=3.0
     dj18: Django>=1.8,<1.9
     dj19: Django>=1.9,<1.10
     dj110: Django>=1.10,<1.11
+    postgres: psycopg2>=2.6
```
jrief: Wouldn't it be safer to pin these requirements on major version numbers, say ``psycopg2>=2.6,<2.7``?
Sometimes it happened that we receive a pull request which works in theory, but because of some API changes in the underlying dependency the unit tests fail.

tox.ini
```diff
@@ -19,8 +19,19 @@ commands =
...
+    mysql: mysqlclient>=1.3.9
```
jrief: detto

tox.ini
mvantellingen: Sure. In my defence, I've just copied the versions as specified in the requirements_test.txt :-)

mvantellingen: Although I'm actually wondering if that is the way to go. I think we want to error when treebeard isn't compatible anymore with the latest psycopg2 / mysqlclient version.

jrief: @mvantellingen In the case of DB adaptors you're probably right. Unfortunately with other apps it happened more than once that CI failed, not because of new bugs, but because of an implicit dependency change. Therefore we should following the zen of Python: Better explicit than implicit.

PR #67, opened 2017-06-20
Include documentation in source tarball.

fladi:
This makes the tarballs on PyPi suitable for distribution packages again. The
Debian package would otherwise not be able to provide a documentation package.


jrief:
Thanks!
Shall I release version 4.1.2 in PyPI?

fladi:
That would be great. Otherwise I would have switched downloading the source tarball from github.

PR #77, opened 2017-09-08
_default_manager is now used in admin (not objects)

igerko:


jrief:
when offering a pull request, it would be helpful to know what it fixes or changes.

tabo:
@igerko Thanks for the patch. I know it has been a few years, but if you're still interested in collaborating, please file a ticket explaining the goals here. The patch as it is, is introducing changes to the API which would involve a major release. Thanks!

PR #79, opened 2017-09-29
Adding back requirements_test.txt removed in 2a0e8fb

craigloftus:
The `requirements_test.txt` file was removed in 2a0e8fb but it is still being used in the main `requirements.txt` file.


mvantellingen:
Should be fixed now

PR #83, opened 2017-11-20
Added Django 1.11 to CI environments.

atombrella:
I also saw Python 3.6 mentioned in the release notes, so this is also mentioned and adjusted.


treebeard/tests/test_treebeard.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 """Unit/Functional tests"""
 
-from __future__ import with_statement, unicode_literals
```
atombrella: Python 2.5/2.6 legacy.

PR #84, opened 2017-12-03
Django 2.0 support

mvantellingen:
Add Django 2.0 support


mvantellingen:
I think we should do a 4.2 release based on this PR and then in the future a 5.0 release which drops support for Django  1.8, 1.9 and 1.10. Or perhaps even a 5.0 which drops Python 2 support and leverages the window support in Django 2.0

Any opinions @django-treebeard/core-maintainers  ?

yakky:
As long as we keep 4.2.x open for future releases (as 1.11 will be supported until 2020) I'm ok with releasing 5.0 which drops Django < 2.0 and python2

jrief:
I agree with @mvantellingen and @yakky on this. Dropping Python support is a major backwards incompatibility which shall increase the big version number.

jrief:
@mvantellingen but this PR would trigger only a minor version jump? Shall I release it on PyPI as 4.3.0?

mvantellingen:
@jrief if you want to do a release then that would be great. I have to mention however that I havent tested this changeset in an existing project. Although it should work

PR #87, opened 2017-12-15
Fix for `object has no attribute 'rel'` in django 2.0

mjr27:
In django 2.0 admin area for treebeard-enabled model got broken  because of exception in `result_tree` tag.

The fix was rather simple, however

### DIFF
--- a/treebeard/templatetags/admin_tree.py
+++ b/treebeard/templatetags/admin_tree.py
@@ -83,7 +83,7 @@ def get_result_and_row_class(cl, field_name, result):
             if isinstance(value, (datetime.date, datetime.time)):
                 row_classes.append('nowrap')
         else:
-            if isinstance(f.rel, models.ManyToOneRel):
+            if hasattr(f, 'rel') and isinstance(f.rel, models.ManyToOneRel):
                 field_val = getattr(result, f.name)
                 if field_val is None:
                     result_repr = empty_value_display


likhvar:
Most probably `rel` should be replaced with `remote_field` attribute for Django 2.0 support.

sasha0:
I've opened a new PR - https://github.com/django-treebeard/django-treebeard/pull/102 to address this issue considering feedback above within the test.

philippeluickx:
This PR can be closed then, considering https://github.com/django-treebeard/django-treebeard/pull/102 is merged?

treebeard/templatetags/admin_tree.py
```diff
@@ -83,7 +83,7 @@ def get_result_and_row_class(cl, field_name, result):
             if isinstance(value, (datetime.date, datetime.time)):
                 row_classes.append('nowrap')
         else:
-            if isinstance(f.rel, models.ManyToOneRel):
+            if hasattr(f, 'rel') and isinstance(f.rel, models.ManyToOneRel):
```
jrief: if we do it like this, that code will remain forever.
I would like to see a solution which can be safely be removed whenever we drop support for Django<2.0
Please use the version numbers provided by `django.__init__.py`

solarissmoke:
This was fixed in #102.

PR #93, opened 2018-02-08
Allowing for UUID ids

JamesMaroney:
I ran into some issues using `MP_Node` and overriding the model's `id` field.
I was able to add a new class to the tests and followed the bouncing ball to update the code. Thanks for having awesome tests! Made the updates so much easier.

My main concerns with this update are the setting of a `default=1` for the depth field, and the usage of `._state.adding` from outside the model object.

I hope this isn't too far off base, and looking forward to any feedback.


JamesMaroney:
While the tests pass, my changes have broken something when actually using the branch. I will update this PR when I figure it out.

JamesMaroney:
Ok, I've worked past that issue. So far models (MP_Node) seem to be working just fine in the admin interface. Add/Delete/Edit/Move.

rtibbles:
Is there anything stopping this from being merged? This would be very useful to have as a feature.

PR #94, opened 2018-02-08
Fix "ImportError: cannot import name '__version__'"

jedie:
Install as editable raised into this:
```
  Running setup.py develop for django-treebeard
    Complete output from command .../bin/python3 -c "import setuptools, tokenize;__file__='.../src/django-treebeard/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" develop --no-deps:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File ".../src/django-treebeard/setup.py", line 7, in <module>
        from treebeard import __version__
    ImportError: cannot import name '__version__'
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,10 @@ from __future__ import unicode_literals
 
 import os
 from setuptools import setup, find_packages
-from treebeard import __version__
+try:
+    from treebeard import __version__
+except ImportError as err:
+    __version__="<unknown: %s>" % err
 import codecs
 
 


jrief:
I use a similar pattern all the time, and never experienced any problem with it.
How about doing?

```
import treebeard

...

setup(
    ...
    version=treebeard.__version__,
    ...
)
```

then during the invocation of `setup()` all the modules should be ready.

jedie:
Sorry, i didn't know the problematic situation...

I searched a little bit around and find some simmilar issues, e.g.:

https://github.com/cleder/fastkml/issues/23

and maybe a better solution using **pkg_resources** to get the version here:
https://github.com/cleder/fastkml/pull/24/files

more issues here: https://github.com/search?q=%22ImportError%3A+cannot+import+name+__version__%22&type=Issues&utf8=%E2%9C%93

jrief:
@jedie interesting idea to move the canonical version string into the `setup.py`. A problem with this is, that sometimes 3rd party apps use that string, but if it is installed as editable package (ie. `pip install -e https://github/...git#egg=...`) then having just `dev` might be a problem.

My solution proposal from above, doesn't this work?

jrief:
I have never seen this pattern in any existing Python package and I dislike having a version named ``unknown``. Library version are not "unknown", they might be something in between, say 1.2.3.rc4.

PR #95, opened 2018-02-16
Remove links that doesn't work

EmilStenstrom:


EmilStenstrom:
Hmm... those links are not available in master? I see them here though: http://django-treebeard.readthedocs.io/en/stable/

PR #96, opened 2018-02-22
Update install.rst

andytwoods:
update for modern django


jrief:
Many thanks to the savvy reader.

PR #102, opened 2018-04-02
issues/django2

sasha0:


jrief:
@sasha0 with that fix, can we approve the next version of django-treebeard for Django-2.0?

sasha0:
Sure. We didn't find any other issues on Django 2.0 so far.

jrief:
@sasha0 OK, then I will release a new version.

jrief:
Just released version 4.3 on PyPI.
Thanks for the reminder.

PR #103, opened 2018-04-14
Update supported Django versions in docs

jdufresne:
Support for Django 1.7 was removed in version 4.1.0 (Nov 24, 2016). Update the docs to mention 1.8+ is supported.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -6,7 +6,7 @@ Prerequisites
 -------------
 
 ``django-treebeard`` needs at least **Python 2.7/3.4** to run, and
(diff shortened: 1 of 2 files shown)


docs/source/install.rst
```diff
@@ -6,7 +6,7 @@ Prerequisites
 -------------
 
 ``django-treebeard`` needs at least **Python 2.7/3.4** to run, and
-**Django 1.7 or better**.
+**Django 1.8 or better**.
```
jrief: wouldn't be *later* or *newer* be a better wording?

Issue #81, opened 2017-10-09
Error "path already exists" on add new child


brabant:
An error occurs when creating a new node in the CMS Wagtail "this path already exists." 
I change the line 277 in the file models.py and then work normaly
return self.get_children().order_by("path").reverse()[0]

bkfox:
Hey, I also have got this error; there is also something I don't understand: is not "last()" method used in this case? This allows returning None when the node is a leaf.

patta42:
The same happens here, although I cannot provide steps on how to reproduce this. Interestingly, even adding the `order_by('path')` to the return value of `MP_Node.get_children` does not help, but adding it to in the `get_last_child` method does. I suggest to add it in the `get_first_child` method, too.

Another interesting thing is that when I call the `get_last_child` method of the same node from the shell (opened via `manage.py shell`), the result is correct.

Here is my data:

Wagtail: 1.13.1
Django: 1.11.7
Python: 3.4.6

In my opinion, this is something which definitely requires more research. If this happens on a production server, one cannot add new pages to the corresponding tree, which is a severe problem.

gasman:
Without any steps to reproduce, I don't think it's possible to say whether this is an issue with treebeard, Wagtail or the user's own project.

patta42:
Uhhh, here's the problem, at least for my case. I was too blind to see it, sorry.

Okay, here are the steps to reproduce the error:

Make a wagtail `Page`-model which overrides the `get_children()` method and changes the order of the children.

```
class Testpage ( Page ):
    def get_children( self ):
        return super(Testpage, self).get_children().order_by('title')
```

1. Create a Testpage instance
2, Add a child with title `c` to the Testpage
3. Add a child with title `a` to the Testpage
4. Adding another child will now throw the error.


The reason is very obvious: the `get_last_child()` method of `Node` relies on the unchanged order of the result of its `get_children()` method, which was obviously changed by the model above.

I don't know whether this should be considered a bug or not, I would suggest to add a method `_get_children()` to `Node`, with the underscore indicating that it should not be touched, and then use this method in the `get_last_child()` method (and `get_first_child()` accordingly).

At least I would suggest to update the docs to state that `get_children()` should not be changed.

gasman:
[...]

Issue #65, opened 2017-05-25
Strange out of order root nodes with MP_Node


zbyte64:
Have an issue where the system seems to retrieve the wrong last root node causing the path increment logic to create a conflicting path.

```
>>> Concept.find_problems()
([], [], [], [], [])
>>> [(o, o.path) for o in Concept.get_root_nodes()]
[(<Concept: Law>, '0003'), (<Concept: jurisdiction>, '0002'), (<Concept: identity document>, '0004'), (<Concept: currency>, '0001')]
>>> Concept.get_last_root_node()._inc_path()
'0004' #this path already exists!
>>> pprint(Concept.dump_bulk())
[{'data': OrderedDict([('title', 'Law'),
                       ('slug', 'law'),
                       ('definition',
                        'a system of rules and guidelines which are enforced '
                        'through social institutions to govern behavior'),
                       ('example', ''),
                       ('url_path', 'law'),
  'id': 25},
 {'data': OrderedDict([('title', 'jurisdiction'),
                       ('slug', 'jurisdiction'),
                       ('definition',
                        'the limits or territory within which authority may be '
                        'exercised; the power, right, or authority to '
                        'interpret and apply the law'),
                       ('example', ''),
                       ('url_path', 'jurisdiction'),
  'id': 29},
 {'data': OrderedDict([('title', 'identity document'),
                       ('slug', 'identity-document'),
                       ('definition',
                        'any legal document which may be used to verify '
                        "aspects of a person's identity"),
                       ('example', ''),
                       ('url_path', 'identity-document'),
  'id': 34},
 {'data': OrderedDict([('title', 'currency'),
                       ('slug', 'currency'),
                       ('definition',
                        'medium of exchange value, defined by reference to the '
                        'geographical location of the authorities responsible '
                        'for it'),
                       ('example', ''),
                       ('url_path', 'currency'),
  'id': 23}]
>>> Concept.fix_tree(destructive=True)
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
psycopg2.IntegrityError: duplicate key value violates unique constraint "focus_concept_path_key"
DETAIL:  Key (path)=(0002) already exists.


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<console>", line 1, in <module>
[...]

PR #112, opened 2018-06-21
Sorting on path on necessary queries

philippeluickx:
Fixes some issues and stabilizes the whole MP section. Shouldn't be breaking.

https://github.com/django-treebeard/django-treebeard/issues/81
https://github.com/django-treebeard/django-treebeard/issues/65


jrief:
Thanks.

PR #116, opened 2018-08-01
Pass kwargs to super as kwargs, not args

schwobaseggl:
This will play more nicely with other form classes, that implement `__init__(self, *args, **kwargs)`, e.g. django-parler's `TranslatableModelForm`, where `kwargs.get('instance')` is `None` when called from here as it is passed as `args[8]`. The `BaseModelForm` will still work fine.


jrief:
Does this fix a known bug or did you experience any problem with the current version?

PR #119, opened 2018-08-21
fix drag/drop order not working

TZanke:
variable named incorrectly, drag and drop does not work in admin in current treebeard versions


TZanke:
looks like the testsuite had an error, can you restart it?

PR #131, opened 2019-01-18
Prevent unnecessary database query for get_ancestors() on MP_Tree if the node is a root node

solarissmoke:
Hi! We noticed that calling `get_ancestors()` on a root node will trigger a database query even though the node cannot have ancestors. 

This change makes the behaviour consistent with what [NS_Tree](https://github.com/django-treebeard/django-treebeard/blob/ad44d91d0459750278694c49ed8e9525ff3089ee/treebeard/ns_tree.py#L631-L632) does, and skips the database query if the node is a root node.


jrief:
please add a short description for the changelog

solarissmoke:
Sure - see #133

PR #141, opened 2019-05-30
Updating delete method to respect return value from django

BigE:
Currently the delete() method always returns `None`. This does not support the default Django delete() method default behavior. From the [Django documentation](https://docs.djangoproject.com/en/2.1/ref/models/querysets/#delete), it will
> Performs an SQL delete query on all rows in the QuerySet and returns the number of objects deleted and a dictionary with the number of deletions per object type.

This change, essentially, returns the expected Django result when calling the delete() method, even when there is nothing to delete.

### DIFF
--- a/treebeard/mp_tree.py
+++ b/treebeard/mp_tree.py
@@ -123,7 +123,9 @@ class MP_NodeQuerySet(models.query.QuerySet):
         # ids, and will deal with removing related objects
         if toremove:
             qset = get_result_class(self.model).objects.filter(reduce(operator.or_, toremove))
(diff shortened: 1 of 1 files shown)


solarissmoke:
Makes sense to return a value consistent with what Django would, but if we do it should be done consistently for all the tree implementations, not just `MP_Tree`.

hartwork:
@solarissmoke I agree. Given no reply since 2019-05, should this be closed as obsolete?

tabo:
@BigE Thanks for the patch. I filed #197 to start a discussion on a solution for this. Thanks!

PR #143, opened 2019-07-16
Add migrations with path COLLATE for databases they provides it.

zbohm:
Values in `path` must be sorted by ascii alphabetical order, otherwise module doesn't find the tail and creates duplicate path. Some databases have default collations that sorts in different way. It is necessary to set a proper collation to the column `path`.

 * https://www.postgresql.org/docs/9.1/collation.html
 * https://oracle-base.com/articles/12c/column-level-collation-and-case-insensitive-database-12cr2
 * https://dev.mysql.com/doc/refman/8.0/en/charset-collate.html

### DIFF
--- a/treebeard/migrations/0001_treebeard_path.py
+++ b/treebeard/migrations/0001_treebeard_path.py
@@ -1,0 +1,16 @@
+from django.conf import settings
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    operations = []
+    if settings.DATABASES['default']['ENGINE'] in (
+            'django.db.backends.postgresql_psycopg2', 'django.db.backends.postgresql'):
+        operations.append(migrations.RunSQL(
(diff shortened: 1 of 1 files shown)


solarissmoke:
@zbohm Treebeard doesn't ship with any concrete models, and therefore has no migrations. The migrations you've added are for a `cms` table, which presumably comes from a concrete model you've created as a subclass of one of Treebeard's abstract models. They would have no effect elsewhere.

I don't think this solution will work generically in Treebeard - and is perhaps something that just has to be documented for projects to be aware of.

jacobwegner:
For the record, I hit this on a project and we ended up subclassing `MP_Node` and passing an explicit collation to work around some differences between Postgres in dev / production environments, e.g.:

```python
POSTGRES_COLLATION_C = "C"


class BaseNode(MP_Node):
    # NOTE: This subclass of MP_Node is being used to ensure that
    # the `path` field has consistent sorting.
    path = models.CharField(
        max_length=255, unique=True, db_collation=POSTGRES_COLLATION_C
    )
```

Issue #146, opened 2019-10-13
Allow extending delete methods


virtosubogdan:
**Issue**
Trying to extend the delete operation is not possible without patching. 
Because single node delete operations must turn into queryset deletes, certain overrides will be prevented.

This is an example:
```python
class SafetyQueryset(models.query.QuerySet):
    def delete(self, force_delete=False):
        for obj in self.all():
            obj.delete(force_delete=force_delete)


class SafetyManager(MP_NodeManager):
    _queryset_class = SafetyQueryset

    def get_queryset(self):
        return self._queryset_class(self.model, using=self._db)


class SafetyMixin(models.Model):
    objects = SafetyManager()

    class Meta:
        abstract = True

    def delete(self, force_delete=False):
        if force_delete:
            super(SafetyMixin, self).delete()


class SafeMP_TestNode(SafetyMixin, MP_Node):
    pass

SafeMP_TestNode.add_root().delete(force_delete=True) # will not delete
SafeMP_TestNode.objects.all().delete(force_delete=True) # will not delete
```
This is longer than needed but hopefully it illustrates the problem better.

My original issue was the same as https://github.com/django-treebeard/django-treebeard/issues/109, but I tried to turn into a general delete problem, that would be more fitted for testing inside django-treebeard.

**Proposal**
This code should be valid as per https://django-treebeard.readthedocs.io/en/latest/caveats.html, but I understand that it refers to an edge case.

Passing `*args, **kwargs` would not fix this use case.
A possible option is to add any hook to make treebeard use the original manager for deletes.
I'll open a pull request with tests and proposal fix, but I'm open to implementing this any other way.

jrief:
> Because single node delete operations must turn into queryset deletes, certain overrides will be prevented.

what do you mean with *certain overrides*?

virtosubogdan:
I meant to say that this `force_delete` override, while useful, is a bit weird.
A logging or an "also do this override" would not have this issue.

tabo:
@virtosubogdan Hi. I think your problem is that you are not subclassing the MP_NodeQuerySet as recommended in the [warnings of MP trees](https://django-treebeard.readthedocs.io/en/latest/mp_tree.html#treebeard.mp_tree.MP_Node). That should solve your issues with deletion. Please reopen the ticket if the problem persists.

PR #147, opened 2019-10-13
Proposal. Use default manager to allow extending delete.

virtosubogdan:
Proposal implementation for https://github.com/django-treebeard/django-treebeard/issues/146 improvement


tabo:
This can be solved by implementations properly subclassing MP_NodeQuerySet.

virtosubogdan:
Thanks for replying
I can't see how subclassing MP_NodeQuerySet would fix the delete behavior from the test. Could you please give me a short example ?

PR #153, opened 2020-02-11
Declare support for Django 3 and drop support for EOL Django/Python versions.

### LINKED ISSUE #148: djaango 3.0 supported?

### LINKED ISSUE #152: Deprecation warning due to invalid escape sequences in Python 3.8
Deprecation warnings are raised due to invalid escape sequences in Python 3.8 . Below is a log of the warnings raised during compiling all the python files. Using raw strings or escaping them will fix this issue.

```
find . -iname '*.py'  | xargs -P 4 -I{} python -Walways -m py_compile {}

./treebeard/models.py:23: DeprecationWarning: invalid escape sequence \*
  """
./treebeard/models.py:365: DeprecationWarning: invalid escape sequence \*
  """
./treebeard/models.py:387: DeprecationWarning: invalid escape sequence \*
  """
./treebeard/tests/settings.py:46: DeprecationWarning: invalid escape sequence \S
  'HOST': '(local)\SQL2016',
```

solarissmoke:
Hi!

This PR:

1. Adds declared support for Django 3 and fixes a few minor deprecation warnings that are issued in Django 3. Everything else looks to be working fine.
2. Adds support for Python 3.8.
3. Drops support for Django 2.0 and below. Django 2.0 and 2.1 are both end of life, and Django 1.11 reaches end of life in April 2020. Django [recommends](https://docs.djangoproject.com/en/3.0/releases/3.0/#third-party-library-support-for-older-version-of-django) that all third party libraries should drop support for it now.
4. Drops support for Python 2, since there are no supported versions of Django that support it.
5. Fixes some invalid escape strings that will break in future versions of Python.

Also fixes #148, fixes #152.


jrief:
I agree to dropping support for Django-1.10 and lower. I also agree to drop support for Django-2.0 and 2.1, but why Django-1.11? I can imagine that there are still a few projects around which have been built on the LTS version of 1.11, so dropping it might be too early.

raiderrobert:
I concur. Django 1.11 isn't EOL planned until April this year:
https://www.djangoproject.com/weblog/2015/jun/25/roadmap/

solarissmoke:
Thanks @jrief @raiderrobert for the feedback. I'd ask you to reconsider however regarding Django 1.11. Here is what the [Django 3 release notes](https://docs.djangoproject.com/en/3.0/releases/3.0/#third-party-library-support-for-older-version-of-django) say:

> Following the release of Django 3.0, we suggest that third-party app authors drop support for **all versions of Django prior to 2.2**.

This suggestion is I think based on sound reasoning. Django 1.11 will reach end of life in April - this is *less than two months away*. Thus projects should be well into the process of upgrading. It would certainly be a bad idea to wait until *after* it is no longer supported to upgrade - especially because it is an LTS release. Thus the community should be encouraging all developers to complete that upgrade well in advance.

On this basis I don't think there is any reason not to drop Django 1.11 support in Treebeard - based on the assumption that nobody is going to start a new project using Django 1.11 between now and April. If anyone does need it, then they can simply stick to the current version for another 6 weeks?

solarissmoke:
Just a note that Django 1.11 is now officially end of life.

hartwork:
@solarissmoke could this be rebased against latest `master`? Thank you!

solarissmoke:
@hartwork rebased. Appveyor build is failing because of the known issue that `django-pyodbc-azure` doesn't support Django 2.2 or higher. It seems that it should be allowed to fail until that is resolved.

hartwork:
Hi @solarissmoke , thanks for the rebase. Regarding allowed to fail, either please mark as "allowed to fail" in AppVeyor so that AppVeyor goes back to green on the outside, or drop AppVeyor (because I don't see Django 2.2. support happening at https://github.com/michiya/django-pyodbc-azure/issues/192 and mssql is not even officially supported according to treebeard's readme) or keep AppVeyor at Django 2.1 as the closest working to 2.2 despite end-of-life. Having AppVeyor back to a red cross (and hence commits back to a red cross on `master` after merge) is the worst in my book because people will learn that red crosses are normal and okay which they should not. I know all the other options are not ideal either. Just my two cents. What do you and others think?

solarissmoke:
> Please mark as "allowed to fail" in AppVeyor so that AppVeyor goes back to green on the outside, or drop AppVeyor. 

I think one of the core maintainers of this package needs to make that decision. I'm happy to update the PR with some guidance on which way to go. Seems to me preferable to drop it completely rather than allowing everything to fail.

hartwork:
@jrief what do you think about https://github.com/django-treebeard/django-treebeard/pull/153#issuecomment-654176978 ?

jrief:
Just had a look at the error report from AppVeyor on django-treebeard. No idea what happens there, neither how to fix it.
I'm perfectly OK with this pull request for dropping support for Python-2.7 and Django-1.11, since they reached their EOL.
How can we fix this? I need someone with more experience on Windows and MSSQL.

hartwork:
> Just had a look at the error report from AppVeyor on django-treebeard. No idea what happens there, neither how to fix it.

Error `django.core.exceptions.ImproperlyConfigured: Django 2.2.14 is not supported."` at https://ci.appveyor.com/project/mvantellingen/django-treebeard/builds/33922703/job/be8sfyixm1nxks0n#L140 is from this code in django-pyodbc-azure at https://github.com/michiya/django-pyodbc-azure/blob/a902a956e2a492ec349d4637a618183c14216ec7/sql_server/pyodbc/base.py#L11-L12 . That's ticket https://github.com/michiya/django-pyodbc-azure/issues/192 upstream.

shapiromatron:
Need any help with this PR? It'd be good to get a new release out now that django 3.1 has been released.

solarissmoke:
@jrief I've removed the test config for MSSQL, since those all depended on django-pyodbc-azure which is no longer maintained. Would you be able to remove the appveyor integration from the repo, since it now has nothing to do?

hartwork:
On a sidenote, we'll keep wasting CI cycles on end-of-life rather than recent version that deserve coverage until this is merged, e.g. see https://travis-ci.org/github/django-treebeard/django-treebeard/builds/738325618 for an example.

We can even drop Python 3.5 from here by now if you like, as it's end-of-life already, too.

solarissmoke:
I'm taking the liberty of merging this, as it's blocking a lot of other contributions to this repository. AppVeyor builds will continue to fail - we'll just have to ignore that until someone with permissions to do so can unlink it.

hartwork:
Awesome, great to have that on `master` finally!

PR #158, opened 2020-03-25
Ensure post_save is triggered after parent is updated in add_child

simonschmidt:
This avoids strange behaviours in post_save receivers
since if the parent is in an inconsistent state many types
of queries depending on numchild does not work correctly


simonschmidt:
Rebased to make the tests pass thanks to the CI setup fixes

PR #165, opened 2020-06-26
Implement a non-destructive path-fixing algorithm for MP_Node.fix_tree

gasman:
From [the `MP_Node.fix_tree` docs](https://django-treebeard.readthedocs.io/en/latest/mp_tree.html#treebeard.mp_tree.MP_Node.fix_tree):

> That needs complex in-place tree reordering, not available at the moment (hint: patches are welcome).

It might have taken 11 years, but I took the hint. :-)

This replaces `MP_Node.fix_tree(destructive=True)` with an implementation that non-destructively fixes 'holes' in paths and applies `node_order_by` ordering. It runs after the old `fix_tree` code has fixed up the `depth` and `numchild` values, and works pretty much as you'd expect:

* for each non-leaf node in the tree, fetch the child nodes in `node_order_by` order if specified, or `path` order if not; enumerating this queryset gives us the desired final path positions
* build a lookup table of the nodes' current path positions
* for each node in the queryset:
  * if the node is already in its desired final position, do nothing
  * if the desired slot is vacant, update the path of that node and its descendants to move it into that slot
  * if the desired slot is occupied by another node, move that node to one greater than the previous max position (updating the lookup table accordingly) and move the correct node into the slot

(note: I took the liberty of increasing the alphabet length on the test models. Fixing a tree in-place with no spare characters isn't really a fair challenge!)

### DIFF
--- a/treebeard/mp_tree.py
+++ b/treebeard/mp_tree.py
@@ -6,7 +6,8 @@ from functools import reduce
 
 from django.core import serializers
 from django.db import models, transaction, connection
-from django.db.models import F, Q
+from django.db.models import F, Q, Value
+from django.db.models.functions import Concat, Substr
 from django.utils.translation import gettext_noop as _
 
 from treebeard.numconv import NumConv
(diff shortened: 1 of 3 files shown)


solarissmoke:
@gasman would be grateful if you can rebase this onto master so we can get it merged.

gasman:
@solarissmoke No problem - rebased now.

solarissmoke (approved):
Nice one @gasman . Worth the 11-year wait :P.

tabo:
I came to this closed PR to say: fantastic work @gasman !

PR #166, opened 2020-06-29
Fix quoting in mixed multidatabase configurations

truetug:
treebeard don't respect model db connection settings and does not work if you have config with default mysql and secondary postgres because postgresql can't handle backquotes from mysql

### DIFF
--- a/treebeard/mp_tree.py
+++ b/treebeard/mp_tree.py
@@ -7,7 +7,7 @@ if sys.version_info >= (3, 0):
     from functools import reduce
 
 from django.core import serializers
-from django.db import models, transaction, connection
+from django.db import models, transaction, connections
 from django.db.models import F, Q
 from django.utils.translation import ugettext_noop as _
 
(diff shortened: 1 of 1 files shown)


jrief:
Unless you fix this pull request to run on Travis-CI, I see no chance that it ever will be merged.

truetug:
jrief, with all due respect, it is obviously a problem of travis-ci.  But I'll try to fix it too.

```
psql: could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
The command "psql -c 'CREATE DATABASE travis_ci_test;' -U postgres" failed and exited with 2 during .
```

tabo:
@truetug Thank you very much for the patch. As it is right now, it is incomplete (only supports MP) and expensive, since it has a for loop every time it is needed. I filed #196 to start a conversation about this problem. Thanks!

PR #168, opened 2020-07-01
Fix AppVeyor and Travis CI (fixes #167)

### LINKED ISSUE #167: CI failing on master since 2019-01-23
Hi!

From a quick look it seems that CI has been failing for Git master since 2019-01-23, currently both Travis and AppVeyor, see https://github.com/django-treebeard/django-treebeard/commits/master . Is anyone working on fixing the CI?

Best, Sebastian

hartwork:
Fixes #167

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -110,6 +110,7 @@ before_install:
   - sudo apt-get --yes remove postgresql\*
   - sudo apt-get install -y postgresql-11 postgresql-client-11
   - sudo cp /etc/postgresql/{9.6,11}/main/pg_hba.conf
+  - sudo sed 's,port = 5433,port = 5432,' -i /etc/postgresql/11/main/postgresql.conf
   - sudo service postgresql restart 11
 
 install:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,12 +3,9 @@ services:
 
 environment:
   matrix:
-    - TOXENV: py27-dj18-mssql
-    - TOXENV: py27-dj19-mssql
-    - TOXENV: py27-dj110-mssql
-    - TOXENV: py35-dj19-mssql
-    - TOXENV: py36-dj110-mssql
     - TOXENV: py36-dj111-mssql
+    - TOXENV: py36-dj20-mssql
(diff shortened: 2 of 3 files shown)


jrief:
Thanks so much for your effort.
Just a question of curiosity, what makes AppVeyor better than Travis-CI.

hartwork:
> Thanks so much for your effort.
> Just a question of curiosity, what makes AppVeyor better than Travis-CI.

Happy to help. With regard to AppVeyor, I think there was a time in the past where it was one of the few options offering CI running Windows to cover testing on Windows (and it's used to run Microsoft SQL Server on Windows in here). Personally, I'm rather unsatisfied by now of how unstable Travis is, how their status page pretends that everything is good and green even when it's not, and how their Ubuntu images were stuck in the stone age for way too long. They also sort-of gave up on https://github.com/travis-ci/apt-source-safelist#apt-source-safelist while people in issues overlook it and are not being told: I only found that test now. Maybe having multiple CIs is not the worst, i.e. you get more free resources than with a single one and can cover a greater variety of platforms or cover the same with less runtime, ideally. I think the Travis CI is maybe testing too many combinations right now, maybe that could be thinned out a bit to reduce runtime, but just any idea. If there is no dedicated ticket for that, maybe a simple ticket elaborating the current CI pain points would be good?

jrief:
OK. Never made that experience. I once attempted to migrate one of my other project's unit tests to CircleCI, but abandoned that.
Anyway, would you mind if we put your information from above into the README?

hartwork:
> Anyway, would you mind if we put your information from above into the README?

You mean about why there is both AppVeyor and Travis CI? If you want to have the readme explain it maybe just say that AppVeyor is there to cover support for Microsoft SQL Server for the moment?

jrief:
yes, exactly. If you want to do it, please proceed. Add it either into the README or somewhere in the docs.

hartwork:
I don't see strong value in an addition like that yet so I want to opt out of providing such a pull request myself for now. I may find other things to contribute that I believe in myself, too.

PR #171, opened 2020-08-04
fix typos in get_annotated_list

HugoDelval:
I've tried to generate the html list (`<ul>` / `<li>`) using the template here: https://django-treebeard.readthedocs.io/en/latest/api.html#treebeard.models.Node.get_annotated_list and it doesn't match the tree here: https://django-treebeard.readthedocs.io/en/latest/_images/graphviz-e7e37dc8512ab2c6748e55df49b011e14f28a879.png.

Maybe something like this would be better?

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -376,12 +376,12 @@ API
      .. code-block:: python
 
             [
-                (a,     {'open':True,  'close':[],    'level': 0})
-                (ab,    {'open':True,  'close':[],    'level': 1})
-                (aba,   {'open':True,  'close':[],    'level': 2})
-                (abb,   {'open':False, 'close':[],    'level': 2})
-                (abc,   {'open':False, 'close':[0,1], 'level': 2})
-                (ac,    {'open':False, 'close':[0],   'level': 1})
+                ('a',     {'open':True,  'close':[],    'level': 0}),
+                ('ab',    {'open':True,  'close':[],    'level': 1}),
+                ('aba',   {'open':True,  'close':[],    'level': 2}),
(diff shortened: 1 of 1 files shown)


docs/source/api.rst
```diff
@@ -376,12 +376,12 @@ API
...
-                (ab,    {'open':True,  'close':[],    'level': 1})
-                (aba,   {'open':True,  'close':[],    'level': 2})
-                (abb,   {'open':False, 'close':[],    'level': 2})
-                (abc,   {'open':False, 'close':[0,1], 'level': 2})
-                (ac,    {'open':False, 'close':[0],   'level': 1})
+                ('a',     {'open':True,  'close':[],    'level': 0}),
+                ('ab',    {'open':True,  'close':[],    'level': 1}),
+                ('aba',   {'open':True,  'close':[],    'level': 2}),
+                ('abb',   {'open':False, 'close':[],    'level': 2}),
+                ('abc',   {'open':False, 'close':[0], 'level': 2}),
```
HugoDelval, line 383: Or maybe `'close':[1]` here

tabo:
Thanks for the patch. The documentation is correct, you should interpret the first value in those tuples as variables holding the objects. Strings there would be incorrect.

And your suggestion to change the last close from [0, 1] to just [1] would leave the result unclosed. The output of the example as it is right now would be:

```html
    <ul>
        <li>
            a
            <ul>
                <li>
                    ab
                    <ul>
                        <li>
                            aba
                        </li>
                        <li>
                            abb
                        </li>
                        <li>
                            abc
                        </li>
                    </ul>
                </li>
            </ul>
        </li>
        <li>
            ac
        </li>
    </ul>
```

PR #175, opened 2020-08-27
add dump_bulk method to MP_NodeManager

truongvan:
dump_bulk use values to return chosen fields.


tabo:
There is already a dump_bulk() that serializes a tree or branch, so that it can be passed to load_bulk(). In this implementation, we use django's serializer to make sure the output can be dumped safely to a file. Please open a ticket to discuss changing the approach, and document pros/cons of changing the implementation, but keep in mind that we have a documented API that mentions that the output of dump_bulk can be loaded with load_bulk, so any change would need to keep that behavior.

PR #177, opened 2020-08-30
Fixed #176 -- Made MoveNodeForm save already constructed instance.

### LINKED ISSUE #176: The MoveNodeForm.save() method does not take into account the already constructed model objects.
The `MoveNodeForm.save()` method does not take into account the already [constructed](https://github.com/django/django/blob/0be51d2226fce030ac9ca840535a524f41e9832c/django/forms/models.py#L400) model instances that are ready for saving. Instead, it [re-constructs](https://github.com/django-treebeard/django-treebeard/blob/e3887791f3137c4868ea1895d98b578ec7087701/treebeard/forms.py#L144) the new model instance from `cleaned_data`.

This means that any change in attribute values that may occur later at the model level (as usual, in the [Model.clean](https://docs.djangoproject.com/en/3.1/ref/models/instances/#django.db.models.Model.clean) method) will be ignored for new model instances.

```python
class Article(models.Model):
    ...
    def clean(self):
        # Set the pub_date for published items if it hasn't been set already.
        if self.status == 'published' and self.pub_date is None:
            # The MoveNodeForm.save() method will ignore this change
            self.pub_date = datetime.date.today()
```

alosultan:
This PR should close this issue #176

### DIFF
--- a/treebeard/forms.py
+++ b/treebeard/forms.py
@@ -134,17 +134,13 @@ class MoveNodeForm(forms.ModelForm):
         position_type, reference_node_id = self._clean_cleaned_data()
 
         if self.instance.pk is None:
-            cl_data = {}
-            for field in self.cleaned_data:
-                if not isinstance(self.cleaned_data[field], (list, QuerySet)):
-                    cl_data[field] = self.cleaned_data[field]
             if reference_node_id:
                 reference_node = self._meta.model.objects.get(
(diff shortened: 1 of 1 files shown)


tabo:
I understand the rationale behind this patch and the ticket, but it lacks a test case exposing the problem that is solved by the patch. This will also help us make sure that we don't break this again in the future.

PR #178, opened 2020-08-31
105/increase performance for move node form is slow when using large trees

girumb:


jrief:
It is hard to believe, that after removing the recursion, the functionality remains the same. I would have to test this in my local development environment before merging.

johanneswilm:
@jrief Did you come to a conclusion? We need this functionality added urgently.

jrief:
OK. Still not convinced, but I will push a new release today.
In case this pull request adds a regression, don't be upset if I revert it.

johanneswilm:
That is fine, @jrief . Please take a look at PR 30 https://github.com/django-treebeard/django-treebeard/pull/30

PR #182, opened 2020-10-23
allow working with custom ids

obayemi:
treebeard was expecting the primary key for models to be integer serials

fixed two problems
1) detection of insertion was checking if the pk was None, but with custom ids with default values (as with uuids) the pk is never None as the default is set at model init. changed all checks to use more reliable `model._state.adding` instead.
2) the forms forced the use of integer as reference id, so it triggered an error when sending an uuid. I used the model field's formfield `to_python` as coerce function if applicable.

I also added tests but let me know if I should add more tests.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -36,4 +36,4 @@ setenv =
     postgres: DATABASE_ENGINE=psql
     mysql: DATABASE_ENGINE=mysql
     mssql: DATABASE_ENGINE=mssql
-commands = pytest
+commands = pytest {posargs}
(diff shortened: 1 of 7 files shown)


obayemi:
oh, I just noticed that was already mostly fixed by #93 ... Should i delete it ?
The only real difference is that I kept the TypedChoiceField and used the id fields's formfield as a coersion method and that I don't check for both instance `pk` and `_state.adding`

obayemi:
also set the "root" / reference node to None in the forms for a more reliable behaviour across different types.

PR #183, opened 2020-10-24
Code cleanups for Python 3 and Django > 2.2

solarissmoke:
- Use simplified `super()` calls now that we've dropped Python 2 support.
- Replace use of deprecated `django.conf.urls.url` with `django.urls.path`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,6 +24,7 @@ setup_args = dict(
     include_package_data=True,
     description='Efficient tree implementations for Django',
     long_description=codecs.open(os.path.join(root_dir(), 'README.rst'), encoding='utf-8').read(),
+    python_requires='>=3.6',
     install_requires=['Django>=2.2'],
     tests_require=['pytest'],
     classifiers=[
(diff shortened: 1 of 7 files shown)


hartwork (commented):
From a quick look, this looks good to me.

Probably the right time to add `python_requires='>=3.6'` to `setup.py` as well?

solarissmoke:
> Probably the right time to add `python_requires='>=3.6'` to setup.py as well?

Yep - done.

hartwork (changes requested):
I found one `super` left to fix: https://github.com/django-treebeard/django-treebeard/blob/47e734236614f6637b752cc35c07dbd9c99af471/treebeard/mp_tree.py#L124

solarissmoke:
If you look closely you'll see that the second argument is not `self`, so we need to retain the long-form version - `super()` would run the method on the wrong object.

hartwork:
> If you look closely you'll see that the second argument is not `self`, so we need to retain the long-form version - `super()` would run the method on the wrong object.

You're 100% right, sorry.

hartwork (approved):

PR #185, opened 2020-10-24
Restore dummy AppVeyor config to get CI back to 200% green

hartwork:


hartwork:
@solarissmoke what do you think?

solarissmoke:
Good idea!

PR #190, opened 2021-01-13
Fix PyPI badge URL in README

thibaudcolas:
There was a typo in the previous URL, `https://https://pypi.python.org/pypi/django-treebeard`. Without the typo it redirects to `https://pypi.org/project/django-treebeard/` so I updated the badge to point directly there.


jrief:
Ups – Thanks!

PR #192, opened 2021-01-15
Alternative pks

johanneswilm:
This is a combined PR of #30 #93 and #182. See discussion in #30 for details.


johanneswilm:
Hey,
to all the involved: This is a combined PR of #30 #93 and #182 (plus some very basic linting) to allow alternative PKs in django-treebeard. See discussion in #30 for details.

It includes all the tests of all three PRs. The Appveyor tests are passing, while the Travis tests are not being executed, likely because I don't have write access to this repository. I suggest this PR should be merged, thereby replacing the previous PRs, and a new release be prepared ASAP.

Let me know if anyone needs something additional included in this PR. I am open to accepting patches. Alternatively, just add send in a PR to django-treebeard directly.

@tabo  @jrief @solarissmoke @kolanos @rhunwicks @obayemi @JamesMaroney

treebeard/forms.py
```diff
@@ -99,6 +97,9 @@ def __init__(self, data=None, files=None, auto_id='id_%s', prefix=None,
         # update the '_ref_node_id' choices
         choices = self.mk_dropdown_tree(opts.model, for_node=instance)
         self.declared_fields['_ref_node_id'].choices = choices
+        # use the formfield `to_python` method to coerse the field for custom ids
+        idFormField = opts.model._meta.get_field('id').formfield()
```
rhunwicks, line 101: This creates a dependency on the primary key being called `id`, so we need to change it to get the pk field regardless of its name.

rhunwicks (changes requested):

treebeard/forms.py
johanneswilm: @rhunwicks True, but that dependency is actually quite a lot of places currently. I've created a PR on top of this one that allows for pk columns with alternative names. This might be a more invasive change though, so it may require a major version number update. See https://github.com/django-treebeard/django-treebeard/pull/193

johanneswilm:
@tabo Sure, the reason for that is that the tests included testing for Python 3.9 and Python 3.9 is only supported by the latest version of the appveyor image, which in turn no longer supports mssql 2016, etc. . So one thing led to another and the only way I could get the Python 3.9 tests to run was to also upgrade everything else.

I guess the main problem is just that there were several outdated things in there and this will not be a problem going forward as the package will get more frequent updates and releases.

tabo:
I just installed sql server 2016 last night so I could run the tests 😂 Good work here! And I agree with @rhunwicks about the usage of 'id', so let's get the other PR moving forward too.

tabo:
And to reiterate my point about changes to test infra, I understand the need for updating the test infrastructure and settings, but the main reason to do it separately is that things could be broken BEFORE this PR lands, and the extended tests could tell us that. Also, if the tests failed (which they didn't), by keeping the changes separate we could know for sure that failure didn't come from the testing environments.

johanneswilm:
Yes, so the upgrade to python 3.9 was already part of one of the three PRs
I merged. The corresponding testing infrastructure was just missing.
Under normal circumstances, if we weren't working on a multi-year backlog,
and if I had written the PR by myself, I'd agree and have multiple atomic
single issue PRs.
So who is going to merge this? The other PR is also ready for review, but
that one may require a major version bump as this could break it for some
existing users. Especially renaming `keep_ids` to `keep_pks`. Not sure.

tabo:
It's merged.

johanneswilm:
Thanks @tabo ! We have tested and it is working for all our projects now. Any idea when we can see a release with this included?

obayemi:
Yes !!! thank you a lot @johanneswilm @tabo I didn't have much hope it would get merged on day... Sorry I was locked out of my giithub account for a few days so I didn't get the notification, but you made a very good job! I'll migrate the project where I use my fork to upstream in a few days :+1:

PR #193, opened 2021-01-15
Alternative pk field name

johanneswilm:
Models don't always have their pk in a column labeled 'id'. This addition allows for alternative pk column names.

### DIFF
--- a/treebeard/models.py
+++ b/treebeard/models.py
@@ -61,7 +61,7 @@ class Node(models.Model):
                     pk=node_data[key])
 
     @classmethod
-    def load_bulk(cls, bulk_data, parent=None, keep_ids=False):
+    def load_bulk(cls, bulk_data, parent=None, keep_pks=False):
         """
         Loads a list/dictionary structure to the tree.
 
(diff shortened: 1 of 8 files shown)


tabo:
This would involve a major version change, since it changes the API for dump_bulk and load_bulk. As such, it would need more information in the CHANGES file about the breaking changes.

We could consider having dump_bulk and load_bulk handle both keep_ids and keep_pks, and letting the warnings library show a deprecation warning when keep_ids is used. We wouldn't need to make a breaking release if we follow this path, but would warn users that the change is coming.

treebeard/models.py
```diff
@@ -85,11 +85,11 @@ def load_bulk(cls, bulk_data, parent=None, keep_ids=False):
...
 
-        :param keep_ids:
+        :param keep_pks:
 
-            If enabled, loads the nodes with the same id that are given in the
-            structure. Will error if there are nodes without id info or if the
-            ids are already used.
+            If enabled, loads the nodes with the same pk that are given in the
+            structure. Will error if there are nodes without pk info or if the
+            pks are already used.
```
tabo, line 92: This reads wrong in prose. Can we change it to read "primary key" instead of "pk"?

johanneswilm:
True. Or we could change the name back to keep_ids even though te field
name doesn't have to be id. But maybe that's just confusing?

treebeard/models.py
johanneswilm: Done

johanneswilm:
@tabo After reading a little more of the code, I realized that the term "id" is actually used quite a few other places when it really means "primary key". Maybe all of these variables should eventually change their name to refer to "pk" rather than "id", but I see it would not really improve the situation to just change the name of the `keep_ids` function to `keep_pks`. So I have reverted that change. That should also remove for a major version bump.

tabo:
Perfect. Let's revisit the "id" situation in the future.

Issue #212, opened 2021-02-19
Define policy for backwards incompatible changes


tabo:
The most recent release, 4.5, has been reported as backwards incompatible by django-cms (#210) and wagtail (#211). I'd like to discuss the problem in this ticket and make decisions on actions and policies to define.

Quick background: treebeard is an *OLD* and stable project. As such, it's filled with warts and idiosyncrasies for an era that doesn't exist any more. It is also written for a version of Django that nobody uses today, and while for most new releases of Django, treebeard is either compatible or needs only minor tweaks (thanks to the dedication of Django's maintainers for backwards compatibility), the fact remains that Django is an evolving framework, and treebeard hasn't been evolving with it.

Now to have a certain sense of stability, Treebeard has a very comprehensive test suite (#200), and follows Semver for version changes. A backwards incompatible change is, for us, a change in the public API. The PR being discussed in this ticket, #192, didn't expose any breaking changes when I reviewed it, and in fact it added more tests. This is the reason why this release is not tagged as a major one, or it would have been version 5.0 As discussed in #30, stability of this library is paramount, an opinion shared both by @jrief who has been active maintaining this library for years, and me.

Now, despite our commitment to stability, and our caution and respect to Semver, two of treebeard's biggest users report upgrade pains and backwards incompatibilities. I consider this a failure in our process and our testing, and would like to get the involved parties into a discussion on how to prevent this from happening again.

First, about the problem with migrations added (#211), I'd like to add potential migrations as a backwards incompatible change. For this, we need to add a test to make sure that the tables created in our tests don't change from version to version. Is there a library that does this so we can just use it? As a side note, what triggered this problem was adding a default value to MP_Node.depth. I removed it from the model in my local copy of current master, and all the tests pass, so I think this change can be reverted, unless somebody has a good reason not to? (pinging @johanneswilm and @obayemi for their input). 

Second, about the deepcopy error (#210), I'm going to have to dig a bit more on this, but what occurs to me as a preventive measure is adding a quick sanity test run for projects of certain size that depend on treebeard's stability. Looking at [pypi](https://pypi.org/project/django-treebeard/) I see that we have maintainers from wagtail (@gasman), django-cms (@yakky) and django-oscar (@mvantellingen and @solarissmoke). If there is guidance on how to have a quick sanity test of developer branches of treebeard against these projects, it would help us fulfill our stability expectations.

[...]

PR #213, opened 2021-02-21
Add a test to check for missing migrations in test app models.

solarissmoke:
This will hopefully flag changes to the Treebeard models that
require a migration, to ensure that such changes are (a) evaluated for
their necessity and (b) shipped as part of a major release.

See #212 .


tabo:
Excellent! I’m on my phone right now but I’ll take a closer look when I’m back on the PC. I see mssql tests failed? That seems odd

solarissmoke:
MySQL failures seem unrelated - maybe just need to restart those builds (I don't seem to have permission to do that).

tabo:
It just failed the tests for MSSQL again. Don't know much about them, but it's the second time this PR makes them fail. I don't know much about SQL Server, and even less about Django connecting to it, but if we are supporting it now, we need to sort this out.

solarissmoke:
MSSQL failures are because `django-mssql-backend` [isn't compatible](https://github.com/ESSolutions/django-mssql-backend/pull/86) with Django>3.0 - it's been fixed on master but they've not cut a release.

tabo:
Oh! Can we disable those tests until that's dealt with then?

solarissmoke:
@tabo yes - I've commented those out of the appveyor config for now. Build should pass this time.

solarissmoke:
A general comment on MSSQL - Django itself [does not](https://docs.djangoproject.com/en/3.1/ref/databases/) support MSSQL,  and there really isn't a robust third-party backend available. I don't think Treebeard can actually guarantee that it will work on MSSQL - and on that basis don't think we should actually declare support for it.

tabo:
@solarissmoke Fantastic work on this PR. Thanks! And you make a fair point on MSSQL support, but if the tests are working with it, then it's safe to say that treebeard will work on it. My only concern is having 2 test infra and how slow those MSSQL tests are.

solarissmoke:
@tabo it should be possible to move these tests to Github actions (thus we can drop both Travis and AppVeyor) - I will have a play with that when I have some more time.

PR #214, opened 2021-02-21
Switch to Github Actions from Travis CI.

### LINKED ISSUE #191: Travis test credentials are hardcoded even for local tests
I am trying to set up a local test environment again, but I can't run the tests locally other than sqlite. I am following the instructions in https://django-treebeard.readthedocs.io/en/latest/tests.html where it says that settings can be overriden by env vars, but that is not the case any more. This changed when travis support was added (which is fantastic!), but now the documentation is outdated and the experience for a new possible developer is harder.

Is there a new/better way to override test settings in django test runs? I haven't worked on this for years but treebeard used to allow concurrent test runs, while right now since the database names are hardcoded, test concurrency is impossible.

### LINKED ISSUE #205: Move automation to Github Actions
Travis has [recently changed their pricing model](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing). In the new free tier, they give 10,000 credits, at a rate of 10 credits per minute of linux node usage. This means 1,000 free minutes per month, which sounds like a lot, but when we're [using around 84 minutes of testing per commit](#200), this gives us 12 test runs in a month before we use our allotment.

Even if Travis decided to keep a free tier (doesn't seem like they will), they are limited to 5 concurrent jobs, while GH actions has 20 concurrent jobs in their free tier. This is big for us because currently we're using 48 jobs per commit (see #200).

The built-in secret store would allow us to integrate the workflow to publish docs and release to pypi from actions itself.

solarissmoke:
Fixes #205, fixes #191.

@tabo could you please enable actions on this repository, so that I can get the builds working on this PR? (I'm not sure why else the build wouldn't have been triggered automatically).

## Summary

This adds Github Action test workflows intended to replace the Travis CI build. There is one build for each Python version that we support - inside that we run `tox` with all the other permutations of database and Django version.

I've cleaned up the Travis configuration at the same time, to remove database creation that was doing nothing (Django creates its own test databases with a `_test` suffix added to whatever you define in `DATABASES`).

In theory it should also be possible to drop AppVeyor, but I will leave that for another day.


tabo:
I think actions will be enabled once they get merged into master?

solarissmoke:
No, the actions should run without merging (this from previous experience trying to set them up).  Also I'm going to need to mess around to get them working in the first place - there's zero chance it will work first time. Leaving the travis config in place makes sense - will restore that once I've got the actions working.

solarissmoke:
(If you're not able to find a way to enable actions, then I'll first add a barebones action workflow which does nothing, which we can merge to master).

tabo:
@solarissmoke I just checked on the project settings, and all actions are allowed.

solarissmoke:
@tabo I've got this working now.

tabo:
Fantastic work! Thanks!

Issue #200, opened 2021-01-17
Reconsider the testing matrix


tabo:
Treebeard has a very comprehensive test suite and has a very extensive testing matrix. As of right now, tox is testing 4 versions of Python, 3 versions of Django, and 4 database backends.

The matrix started many, many years ago when Django was a young framework that [changed frequently and aggressively](https://code.djangoproject.com/wiki/RemovingTheMagic), and since we make use of very optimized raw sql in the library, we had to make sure it worked across DBMSs. When Django added Python 3 support, we added another dimension to the matrix and it has been like this for more than 10 years.

Do we need this testing rigor in 2021? Python seems to be much more conservative these days. What about Django? I would imagine they are very careful now being a established and mature framework. Running the whole test suite in tox takes me around 5 minutes, and I see it takes travis and appveyor around 15 minutes per commit. Are Python and/or Django changing in a pace that merits this testing?

johanneswilm:
I agree on Python testing. 2 versions should be enough - For example the latest and the oldest hat is supported. Django does still change. I need to update my apps with every new version.

tabo:
I was mistaken about travis' run time. It actually takes around an hour: https://travis-ci.org/github/django-treebeard/django-treebeard/builds/755017104

solarissmoke:
The Travis run time is because they have been deliberately throttling builds on open source projects. We switched django-oscar to Github Actions and the same test matrix that took over 2 hours on Travis now runs in ~5 minutes. This isn't to say that the current matrix is necessary - just that the build time can be slashed regardless.

solarissmoke:
The current matrix on Github actions takes less than 3 minutes in total, so I think this is a non-issue now.

PR #217, opened 2021-02-25
Declare support for Python 3.9

solarissmoke:
- Add Python 3.9 to the supported Python versions in setup.py (this is needed to make the package available for that version of Python from PyPi).
- Fix up documentation for running tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -114,7 +114,7 @@ before_install:
   - sudo service postgresql restart 11
 
 install:
-  - pip install tox codecov
+  - pip install tox
 
 script: tox
 
--- a/docs/source/tests.rst
+++ b/docs/source/tests.rst
@@ -24,12 +24,13 @@ You can use all the features and plugins of pytest this way.
 By default the test suite will run using a sqlite3 database in RAM, but you can
 change this setting environment variables:
 
-.. option:: DATABASE_ENGINE
-.. option:: DATABASE_NAME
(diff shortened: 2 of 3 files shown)


tabo:
While I agree that the testing matrix is already pretty big (see #200), I'd rather discuss other alternatives rather than dropping support for a currently supported version of Django in favor of a development version. I think for now, we should just add a quick sanity check for django 3.2, something like a single `py39-dj32-sqlite` job.

solarissmoke:
@tabo that's a little tricky to do with the Github actions setup - also those builds run relatively quickly so it wouldn't be too much of a problem just to add Django 3.2 to the matrix there. What it really slows down is Travis - but since we're dropping that soon I suggest we just add 3.2 to the matrix (and leave 3.0 intact)?

tabo:
What about testing dev versions of Django in a separate action? I was planning of doing this to test treebeard against other projects as sanity checks (#212 discussion). In this case, Django 3.2 is as of right now, unsupported, so I don't see why we would need to add a full matrix of testing to it. What we need is just a sanity check to warn us that treebeard may break with the next version of Django.

solarissmoke:
@tabo OK. For now I have amended this PR only to declare Python 3.9 support on PyPi, and fix up the documentation in a few places. Can look at adding a separate action for dev versions of Django separately later.

tabo:
Fantastic. Thanks!

PR #224, opened 2021-06-30
Adding django3.2 in tox matrix.

awais786:
Adding django3.2 in tox matrix.


tox.ini
```diff
@@ -25,6 +25,7 @@ deps =
     dj22: Django>=2.2,<3.0
     dj30: Django>=3.0,<3.1
     dj31: Django>=3.1,<3.2
+    dj31: Django>=3.2,<3.3
```
solarissmoke, line 28: This should be `dj32`

solarissmoke (changes requested):
Thanks! Please also add django 3.2 to the [Github workflow](https://github.com/django-treebeard/django-treebeard/blob/master/.github/workflows/test.yml#L54), and to the supported versions in the readme.

tox.ini
awais786: @solarissmoke please review.

Issue #220, opened 2021-04-20
add_child bug


waro163:
when I use the add_child function, some errors lead to creating child failure, but root nodes still increase its numchild field.

```python
root = Student.add_root(name="sam", age=18)

root.add_child(name="xxx") #no age field, create child failure

root.add_child(name='xxx', age=15) #error
```

AttributeError: 'NoneType' object has no attribute '_inc_path'

please fix it, thanks

gasman:
This error can be prevented by wrapping the `add_child` call in a transaction, and the [Materialized Path trees](https://django-treebeard.readthedocs.io/en/latest/mp_tree.html) page in the docs does advise this in multiple places:

> This has the advantage of needing very simple and fast queries, at the risk of inconsistency because of the denormalization of parent/child foreign keys. This can be prevented with transactions.

> Checks for problems in the tree structure, problems can occur when: your code breaks and you get incomplete transactions (always use transactions!)

> The problems this method solves are: Nodes with an incorrect depth or numchild values due to incorrect code and lack of database transactions.

If we are indeed still maintaining that this is the responsibility of the developer, then this could arguably be closed as 'wontfix'. (However, there's also an argument that treebeard _could_ add its own transactions here... has that been considered / rejected previously? Would there be any performance implications from doing so?)

koutoftimer:
@gasman IMHO it is way more convenient to leave it open. I was about to create new issue right before I have found this one.

Issue #140, opened 2019-05-21
error with self.node.get_last_child()._inc_path()


m-pasha:
I get error with this line code
_treebeard/mp_tree.py:387_
`newobj.path = self.node.get_last_child()._inc_path()`

**`'NoneType' object has no attribute '_inc_path'`**

self have parent object page and in kwargs is instance.

`backend_1  | Traceback (most recent call last):
backend_1  |   File "/usr/local/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
backend_1  |     response = get_response(request)
backend_1  |   File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 115, in _get_response
backend_1  |     response = self.process_exception_by_middleware(e, request)
backend_1  |   File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 113, in _get_response
backend_1  |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
backend_1  |   File "/usr/local/lib/python3.6/site-packages/django/views/decorators/cache.py", line 31, in _cache_controlled
backend_1  |     response = viewfunc(request, *args, **kw)
backend_1  |   File "/usr/local/lib/python3.6/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
backend_1  |     return view_func(request, *args, **kwargs)
backend_1  |   File "/opt/project/pv/apps/wagtailoverrides/views/pages.py", line 246, in create
backend_1  |     parent_page.add_child(instance=page)
backend_1  |   File "/usr/local/lib/python3.6/site-packages/treebeard/mp_tree.py", line 1013, in add_child
backend_1  |     return MP_AddChildHandler(self, **kwargs).process()
backend_1  |   File "/usr/local/lib/python3.6/site-packages/treebeard/mp_tree.py", line 387, in process
backend_1  |     newobj.path = self.node.get_last_child()._inc_path()
backend_1  | AttributeError: 'NoneType' object has no attribute '_inc_path'
backend_1  | [21/May/2019 09:03:28] "POST /admin/pages/new/article/articlepage/5/ HTTP/1.1" 500 110440`

tech info:
python 3.6
Django 2.2.1
wagtail 2.5.1
django-treebeard 4.3
<img width="661" alt="image" src="https://user-images.githubusercontent.com/18262789/58092959-ab106000-7bcd-11e9-9089-d7f228fe8131.png">

[...]

PR #234, opened 2021-10-02
solving add_child error

hazho:
solved issues:
#140
#220

### DIFF
--- a/treebeard/mp_tree.py
+++ b/treebeard/mp_tree.py
@@ -373,8 +373,11 @@ class MP_AddChildHandler(MP_AddHandler):
                       ' increasing the path.max_length property'
                       ' and UPDATE your database'))
         else:
-            # adding the new child as the last one
-            newobj.path = self.node.get_last_child()._inc_path()
+            # adding the new child as the last one after checking if there is a Last Child
+            if self.node.get_last_child():
+                newobj.path = self.node.get_last_child()._inc_path()
+            else:
+                newobj.path = self.node.path + '0001'
 
         get_result_class(self.node_cls).objects.filter(
             path=self.node.path).update(numchild=F('numchild')+1)
--- a/treebeard/mp_tree.py
+++ b/treebeard/mp_tree.py
@@ -374,7 +374,7 @@ class MP_AddChildHandler(MP_AddHandler):
                       ' and UPDATE your database'))
         else:
             # adding the new child as the last one after checking if there is a Last Child
-            last_child = self.node.get_last_child():
+            last_child = self.node.get_last_child()
             newobj.path = (
                 last_child._inc_path()
                 if last_child is not None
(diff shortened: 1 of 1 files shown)


treebeard/mp_tree.py
```diff
@@ -373,8 +373,11 @@ def process(self):
                       ' increasing the path.max_length property'
                       ' and UPDATE your database'))
         else:
-            # adding the new child as the last one
-            newobj.path = self.node.get_last_child()._inc_path()
+            # adding the new child as the last one after checking if there is a Last Child
+            if self.node.get_last_child():
```
graingert, line 377: No need to call get_last_child twice

hazho: Had been called Only once, is there another call of it that I am not aware of?

hazho: my changes can become like:
newobj.path = [self.node.get_last_child()._inc_path() if self.node.get_last_child() else self.node.path + '0001']

graingert: That's still two calls...

treebeard/mp_tree.py
```diff
@@ -373,8 +373,11 @@ def process(self):
...
+                newobj.path = self.node.get_last_child()._inc_path()
+            else:
+                newobj.path = self.node.path + '0001'
```
graingert, line 380:
```suggestion
            last_child = self.node.get_last_child():
            if last_child is not None:
                newobj.path = last_child._inc_path()
            else:
                newobj.path = self.node.path + '0001'
```

graingert: but you can actually just use the absence of a child node to determine if a node is a leaf node: https://github.com/django-treebeard/django-treebeard/pull/235/files#diff-35501ef525349cd39e4713d1a9f64a249fa4fbd31d875513e3a15e65988701a2R365

treebeard/mp_tree.py
hazho: I'm not aware how to make it one, please, feel free to modify

treebeard/mp_tree.py
```diff
@@ -373,8 +373,13 @@ def process(self):
                       ' increasing the path.max_length property'
                       ' and UPDATE your database'))
         else:
-            # adding the new child as the last one
-            newobj.path = self.node.get_last_child()._inc_path()
+            # adding the new child as the last one after checking if there is a Last Child
+            last_child = self.node.get_last_child():
```
graingert, line 377:
```suggestion
            last_child = self.node.get_last_child()
```

treebeard/mp_tree.py
```diff
@@ -373,8 +373,13 @@ def process(self):
...
+            newobj.path = (
+                last_child._inc_path()
+                if last_child is not None
+                else self.node.path + '0001'
```
graingert, line 381: I'm not sure where you got this path format from - it should be via `self.node_cls._get_path(self.node.path, newobj.depth, 1)` right?

graingert: it looks like what's happening here is the `self.numchild` attribute has been updated but the database doesn't have the child there

PR #235, opened 2021-10-02
add first child if there's no last child

graingert:


treebeard/mp_tree.py
```diff
@@ -374,7 +375,7 @@ def process(self):
                       ' and UPDATE your database'))
         else:
             # adding the new child as the last one
-            newobj.path = self.node.get_last_child()._inc_path()
```
graingert, line 377: it looks like what's happening here is the `self.numchild` attribute has been updated but the database doesn't have the child there

hazho: thanks to me (;

graingert: ?

hazho (commented):
I tested in my end, your changes are currently better than mine, but both needs more tests to be done on

graingert:
> I tested in my end, your changes are currently better than mine, but both needs more tests to be done on

Thanks, do you have a repeat for this crash it seems to be intermittently happening for me

hazho:
> Thanks, do you have a repeat for this crash it seems to be intermittently happening for me

it been more than 2 years I'm working on it without any issues till before two days, I'm using it with wagtail-cms, I was getting this error only when I was initializing my **first time data** using django-shell (based on Python shell), the rest of treebeard nodes are being created by website's editors/authors who use buttons that being controlled by views.

wanglophile:
Any update on merging this? Seems like a relatively straight-forward fix.

janbaykara:
@jrief seeing as you were the last contributor to merge into `main`, would you mind having a look? The bugs this PR addresses are quite frequent in my experience of Wagtail development.

solarissmoke:
I'm happy to review this, but I do think it will need tests to cover the change and illustrate the issue it is fixing.

PR #257, opened 2022-09-22
fix issue related to adding roots with multiple proxy models

rogue26:
I ran into an issue creating roots when working with multiple proxy models of an `MP_Node `model. `MP_AddRootHandler `was setting `path` based on the class instead of the base class, which can result in `path` being set to an existing value (e.g. not seeing `path=0001` from another proxy model off of the same `MP_Node` model). Adding both `cls `and `base_class `to the definition of `MP_AddRootHandler `and using them appropriately fixes the issue.


solarissmoke:
Please could you add a test to illustrate the problem being fixed, and make sure we don't regress in future?

PR #261, opened 2022-12-24
Drop dependency on jQuery UI.

### LINKED ISSUE #253: jquery-ui vulnerabilities
https://www.tenable.com/plugins/was/113042
https://www.cvedetails.com/vulnerability-list/vendor_id-6538/product_id-31126/Jquery-Jquery-Ui.html

All versions below 1.13 are affected (!!!) and you are using 1.8.5 ;)

It would be great if you could update and negate these vulnerabilities! 

XSS and Exec Code XSS possibilities ;)

solarissmoke:
Drops Treebeard's dependency on jQuery UI. We were only using it for one utility (`disableSelection`), and including the whole library exposes us to all the security issues that come with it. This replaces jQuery UI with our own implementation of the same functionality.

Fixes  #253.


gasman (approved):
Looks good to me - code changes make sense, and I've confirmed that the new `disableSelection` / `enableSelection` helper functions correctly prevent the dragged item's text from being selected.

PR #263, opened 2022-12-25
Update node delete() methods to match Django's delete API.

### LINKED ISSUE #61: Node.delete doesn't support 'using' kwarg
Django model instance ```delete``` method supports ```using``` kwarg since at least Django 1.7. 
treebeard's ```Node.delete``` should support it as well, but then ```delete``` method on node querysets might have to be fixed as well.

### LINKED ISSUE #160: MPNodeQuerySet.delete method does not have 'queryset_only' attribute set
Django's `QuerySet.delete` method has the attribute `queryset_only = True` [(source)][1]. This means that when `Manager.from_queryset(SomeQuerySet)` is used, the `delete()` method is not copied as a manager method[(source)][2].

Django documents this behaviour as a sort of safety feature: https://docs.djangoproject.com/en/3.0/topics/db/queries/#deleting-objects

Treebeard's MPNodeQuerySet does not have this attribute set. This means that calling `Manager.from_queryset(MPNodeQuerySet)` the resulting manager gets a `.delete()` method.

This is written from the perspective of the Wagtail project, whose `PageQuerySet` inherits from `MPNodeQuerySet`. In Wagtail it is possible to call `Page.objects.delete()`, whereas for any vanilla Django models, even within a Wagtail project, the response would be get:

```
>>> MyModel.objects.delete()
*** AttributeError: 'Manager' object has no attribute 'delete'
```

[1]: https://github.com/django/django/blob/b9cf764be62e77b4777b3a75ec256f6209a57671/django/db/models/query.py#L667
[2]: https://github.com/django/django/blob/b9cf764be62e77b4777b3a75ec256f6209a57671/django/db/models/manager.py#L92

I consider this a bug, because it deviates from the Django documentation, and is not itself documented.

Possible resolutions:

- decide it's not a bug
- document it
- add `.queryset_only` attribute to the `MPNodeQuerySet.delete` method

I have opened the related Wagtail issue https://github.com/wagtail/wagtail/issues/5937

### LINKED ISSUE #197: The delete method should bring up to date with current Django standards
As reported in #141, Django 2+ returns a value on [delete()](https://docs.djangoproject.com/en/3.1/ref/models/querysets/#delete). 

We need to update this behavior, which would be a breaking change since the[ API doesn't mention a return value at the moment](https://django-treebeard.readthedocs.io/en/latest/api.html#treebeard.models.Node.delete).

There is an  first attempt at an implementation in #141 but it is only for MP trees and it has no tests.

Necessary tasks:

- [ ] Edit all the tests that call delete to assert they return an expected value
- [ ] Update MP delete
- [ ] Update NS delete
- [ ] Update AL delete
- [ ] Update delete documentation
- [ ] Prepare a new major release

solarissmoke:
- Return the number of objects deleted and a dictionary with the number of deletions per object type, consistent with what Django does.
- Accept aribitrary arguments to the delete() method to support overridden models and Django's own `using` and `keep_parents` arguments.
- Set `alters_data` and `queryset_only` attributes on delete methods, to prevent them being used in an unwanted context.

I wonder whether this is enough of an API change that we need to do a major release rather than minor - open to input on this.

Fixes #197, fixes #61, fixes #160.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,18 @@
+Release 4.6 (in development)
+----------------------------
+
+* Drop support for Django 3.1 and lower.
+* Add support for Django 4.0 and 4.1.
+* Drop support for Python 3.7 and lower.
+* Add support for Python 3.10 and Python 3.11.
+* Change the return value of `delete()` for all node classes to be consistent with Django,
+  and return a tuple of the number of objects deleted and a dictionary with the number of 
+  deletions per object type.
+* Change the `delete()` methods for all node classes to accept arbitrary positional and 
+  keyword arguments which are passed to the parent method.
+* Set `alters_data` and `queryset_only` attributes on the `delete()` methods for all node classes
+  to prevent them being used in an unwanted context (e.g., in Django templates).
+
 Release 4.5.1 (Feb 22, 2021)
 ----------------------------
 
--- a/treebeard/models.py
+++ b/treebeard/models.py
@@ -499,9 +499,12 @@ class Node(models.Model):
         """
         raise NotImplementedError
 
-    def delete(self):
+    def delete(self, *args, **kwargs):
         """Removes a node and all it's descendants."""
-        self.__class__.objects.filter(pk=self.pk).delete()
+        return self.__class__.objects.filter(pk=self.pk).delete(*args, **kwargs)
+
+    delete.alters_data = True
+    delete.queryset_only = True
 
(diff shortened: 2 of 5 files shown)


gasman (approved):
All looks good to me.

I don't think anything here calls for a major version bump... while you could make an argument that someone might have taken advantage of the lack of `alters_data` flag to intentionally call `delete` from template code, I don't think it's controversial to say that permitting that (in contrast to default Django behaviour) was a bug.

Issue #268, opened 2023-02-02
Calling `MP_Node.get_descendants` on a leaf node makes unnecessary db calls


Hargre:
Hi! I just noticed that calling [`MP_Node.get_descendants()`](https://github.com/django-treebeard/django-treebeard/blob/master/treebeard/mp_tree.py#L1025) on a leaf node makes a db query that won't really do much:

![image](https://user-images.githubusercontent.com/20994216/216424975-ba367175-edc4-495a-93f9-184153b64370.png)

This can build up quickly if you're iterating over many records. 

I avoided it by simply checking if the current node is a leaf before trying to call `get_descendants` - which is similar to what [NS Tree `get_descendants`](https://github.com/django-treebeard/django-treebeard/blob/master/treebeard/ns_tree.py#L631) and [MP_Node's `get_children`](https://github.com/django-treebeard/django-treebeard/blob/master/treebeard/mp_tree.py#L1004) already do. I would be happy to submit a Pull request making the change - just thought I would check before if there is anything I didn't consider.

solarissmoke:
Hey @Hargre , a PR for this would be very welcome. I don't see any reason why we shouldn't short-circuit the query if the node is a leaf, same was as `get_children()` does.

PR #269, opened 2023-02-03
Stop MP_Node.get_descendants from querying the db on leaf nodes

Hargre:
Fixes https://github.com/django-treebeard/django-treebeard/issues/268 and adds a test case.


solarissmoke (approved):
Thanks!

PR #287, opened 2023-07-28
Add `include_self=False` to `get_descendants()` in `al_tree.py`, `mp_tree.py` and `ns_tree.py`

hyperkai:
I added `include_self=False` to `get_descendants()` as the 2nd argument so that we can choose to include the node itself or not. *I switched from Django-mptt to Django-treebeard because Django-mptt is currently unmaintained but Django-treebeard doesn't have `include_self=False` in `get_descendants()` which Django-mptt has. `include_self=False` is really useful.

### DIFF
--- a/treebeard/al_tree.py
+++ b/treebeard/al_tree.py
@@ -258,11 +258,13 @@ class AL_Node(Node):
         cls._get_tree_recursively(results, parent, depth)
         return results
 
-    def get_descendants(self):
+    def get_descendants(self, include_self=False):
         """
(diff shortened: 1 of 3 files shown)


solarissmoke:
Hey there, veyr happy to incorporate this change but we will need to add some tests to verify the output. Would you be able to do that?

Zerotask:
@hyperkai we're currently also migrating from django-mptt to django-treebeard and also stumbled upon this. Would you be so kind to add tests for your changes so this PR can be merged?

PR #290, opened 2023-08-22
fix mp tree move method

Vaipik:
In the current implementation, when a user attempts to move a node to a parent node that has no children, they encounter an `AttributeError: 'NoneType' object has no attribute is_descendant_of'`. This error arises because currently, the parent node is replaced with its last child, which is `None` if it doesn't have any children: `self.target = self.target.get_last_child()`.

This pull request aims to rectify this issue by adding a check to see if the `get_last_child()` is `None` before replacing `self.target`. This condition ensures that `self.target` is only reassigned if `self.target.get_last_child() is not None`, thereby eliminating the `AttributeError`.

### DIFF
--- a/treebeard/mp_tree.py
+++ b/treebeard/mp_tree.py
@@ -552,7 +552,9 @@ class MP_MoveHandler(MP_ComplexAddMoveHandler):
                 self.pos = 'first-sibling'
                 siblings = get_result_class(self.node_cls).objects.none()
             else:
-                self.target = self.target.get_last_child()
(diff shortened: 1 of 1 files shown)


solarissmoke:
Would you be able to add a test that covers the issue that this change is trying to fix?

solarissmoke:
Looking at this more closely, I don't think the proposed fix has any effect. The block right above the proposed edit performs an `is_leaf()` check, which should handle any target that has no children. I suspect the underlying issue was something else - would be best to produce a failing test if you still have issues.

PR #302, opened 2024-08-19
Fix KeyError in MP_Node.dump_bulk if ordering differs from depth, path.

### LINKED ISSUE #219: sqlite / postgres different treatment for dump_bulk?
Environment:
1. Django: 3.1.7
2. Python 3.9.1
3. Treebeard: 4.5.1

Using *sqlite* as a test case I load key values from a yaml file like so:
```python
results: list[tuple] = self.get_data_from_file(f"{f}.yaml")
first_topic_name: str = results[0][0]
root = Topic.add_root(name=first_topic_name)
for parent, child in results:
    p = create_parent_node(parent, root.path)
    p.add_child(name=child)
```
This creates objects in the database and when performing `dump_bulk`, correctly results in the "data-children" combo.

Applying the same mechanism in *postgresql* (I copied the same django app), I'm able to create the objects in the database just fine... but when performing `dump_bulk`, I get a KeyError:
```
----> 1 Topic.dump_bulk(Topic.get_root_nodes()[0])
~/Code/lawdb/.venv/lib/python3.9/site-packages/treebeard/mp_tree.py in dump_bulk(cls, parent, keep_ids)
    653             else:
    654                 parentpath = cls._get_basepath(path, depth - 1)
--> 655                 parentobj = lnk[parentpath]
    656                 if 'children' not in parentobj:
    657                     parentobj['children'] = []
KeyError: '000100020004'
 ```
 
Have no idea what's causing the discrepancy. Tried searching but the closest related problem I could find dealt with [deleted items and broken trees](https://github.com/django-treebeard/django-treebeard/issues/69#issuecomment-355347616). But no deletions were made in the scenario above.

### PR DESCRIPTION
**Pull request by mvaled**
Fixes #219.

The qset needed to be ordered to ensure parents are serialized before the children.  Otherwise a child might come first and the `lnk` would not have the parent and raise a KeyError, like:

```
Traceback (...):
  ...
  File "tree/models/level.py", line 309, in dump_bulk
    result = super().dump_bulk(parent, keep_ids=keep_ids)
  File "treebeard/mp_tree.py", line 661, in dump_bulk
    parentobj = lnk[parentpath]
KeyError: '00IW'
```

### DIFF
--- a/treebeard/mp_tree.py
+++ b/treebeard/mp_tree.py
@@ -79,7 +79,7 @@ class MP_NodeQuerySet(models.query.QuerySet):
         Custom delete method, will remove all descendant nodes to ensure a
         consistent tree (no orphans)
 
-        :returns: tuple of the number of objects deleted and a dictionary 
+        :returns: tuple of the number of objects deleted and a dictionary
                   with the number of deletions per object type
         """
         # we'll have to manually run through all the nodes that are going
@@ -631,7 +631,7 @@ class MP_Node(Node):
         # Because of fix_tree, this method assumes that the depth
         # and numchild properties in the nodes can be incorrect,
         # so no helper methods are used
-        qset = cls._get_serializable_model().objects.all()
+        qset = cls._get_serializable_model().objects.all().order_by("depth", "path")
         if parent:
             qset = qset.filter(path__startswith=parent.path)
         ret, lnk = [], {}


solarissmoke:
Hey @mvaled thanks for this. It would be great if you are able to add a test that would fail without this change, just so we can avoid regressions in future.

mvaled:
@solarissmoke 

I managed to add a test in #303.  The root cause is custom managers obtained with `MP_NodeQuerySet.as_manager()` (or subclasses).

I will update that PR #303 to add the fix there as well.

PR #306, opened 2024-09-28
remove MP_Node.gap ClassVar

john-parton:
There was an integer ClassVar on MP_Node called `gap`. It doesn't do anything and is completely undocumented, from what I can tell.

Perhaps there was some intention to leaves gaps in the materialized path so that efficient insertions could be done without rewriting the tree?


solarissmoke:
You're right, it's never been used [since it was added](https://github.com/django-treebeard/django-treebeard/commit/3e86d21390cbd05c9f63726965d27b1f37538804). Thanks for spotting!

PR #317, opened 2025-09-05
Clear cached attributes on refresh_from_db()

bmispelon:
When working with Wagtail (which uses treebeard's `MP_Node` quite extensively) I got annoyed that Django's builtin `refresh_from_db()` didn't always work: sometimes, `get_parent()` would report the wrong parent for a model instance, even after that instance had been refreshed.

I dug a little in the codebase and found that it wasn't too hard to fix so here's a PR. Let me know if you'd like me to open an issue also.

Thanks!

### DIFF
--- a/treebeard/al_tree.py
+++ b/treebeard/al_tree.py
@@ -48,6 +48,11 @@ class AL_Node(Node):
     objects = AL_NodeManager()
     node_order_by = None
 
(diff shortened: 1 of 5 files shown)


solarissmoke:
Thanks @bmispelon !

Issue #334, opened 2025-12-17
Async support


evilmonkey19:
Hiii,

I'm currently trying to migrate several projects from sync to async. Is it possible to add async support to the project?

Thanks

solarissmoke:
@evilmonkey19 it's certainly possible, just needs someone who is able to put a bit of time into adding the necessary wrapper functions. A PR is most welcome.

solarissmoke:
Hey @evilmonkey19 thanks for taking the time to look into this. Please give me a few days to reflect on these options before I share an opinion - also a chance for any others who are interested to weigh in.

solarissmoke:
@evilmonkey19 gave this some thought, and I think that it is preferable to go with the approach in your current PR (i.e., adding async methods to the existing classes instead of separate async/sync modules).

The option of making the whole codebase async and then having a sync port would require two separate model classes for all the base models (i.e., `MP_Node` and `AsyncMP_Node`, etc). This means that a project using Treebeard would have to choose between one or the other - and they would be forced to use _only_ the sync methods, or the async ones. This is not consistent with the way Django has implemented async support: Django models expose sync and async methods on the same class, allowing you to use both in any given context. 

I have seen projects which do in fact use sync and async code in different parts of the application (e.g., views vs websocket handlers) - which would be much more difficult if we split things into separate modules. 

I can see the appeal of the approach used by libraries like psycopg3 and elasticsearch-py, but I think the key difference here is that Treebeard extends Django's capabilities and we should follow the patterns used by Django unless there is a compelling reason not to. On that basis I think we should add async methods to the existing classes.

evilmonkey19:
Thanks for giving it a thought! I see your point and I agree. I believe, thought that probably a mixed approach could be possible. Probably, we can develop internally as async and have some way to convert it into sync internally. To be specific, let's take this example:

```python
class MyClass:
  def save(self, *args, **kwargs):
      # sync actions
 
  async def asave(self, *args, **kwargs):
      # async actions
```

We could develop it as:
```python
class MyClass:
  async def save(self, *args, **kwargs):
    # async + sync actions based on if True / False and async
```

So the transpiler could convert into the:

```python
class MyClass:
  def save(self, *args, **kwargs):
      # sync actions
 
  async def asave(self, *args, **kwargs):
      # async actions
```

[...]

PR #335, opened 2026-01-08
[WIP] Async Support

evilmonkey19:
Hi,

I'm adding support for ``async`` (issue #334). The strategy this PR follows is the following:

- Try to avoid as much as possible the ``sync_to_async`` while being somehow pragmatic. I suggest to have full async support as using this helpers is tricky. However, as I want to it to be ready, I use sometimes, when their isn't any other way or I would need much more time to do it.
- I added the prefix ``a`` for the async functions. So for instance, ``get_parent()`` has its async as ``aget_parent``. I tried to follows the naming convention used by django.
- I'm aware that the support sometimes doesn't really use the full async benefits. Hopefully in the future will do. This can be noticed in several places where everything is parsed to a list instead of using queries. In the sync version it doesn't show a lot because the ``.all()`` hides it. 
- There are 2 tests: sync and async. The async one is the same as the sync but for async. **The sync tests have been not been modified**.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,7 @@ dependencies = [
 test = [
     "pytest-django>=4.0,<5.0",
     "pytest-pythonpath>=0.7,<1.0",
+    "pytest-asyncio>=1.3.0",
 ]
 
 [tool.ruff]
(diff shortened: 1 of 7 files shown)


solarissmoke:
Hey @evilmonkey19 , guessing you have got busy with other things, so I am going to close this PR for now. The code underneath has changed a lot, so there are a lot of conflicts which mean it would probably require a fresh attempt anyway.

Happy to reopen in future if/when there is interest again.

PR #339, opened 2026-01-21
Simplify move queries for MySQL

### LINKED ISSUE #199: Remove mysql-specific slow workaround
[This comment](https://github.com/django-treebeard/django-treebeard/blob/master/treebeard/mp_tree.py#L298):

```
            # when using mysql, this won't update the depth and it has to be
            # done in another query
            # doesn't even work with sql_mode='ANSI,TRADITIONAL'
            # TODO: FIND OUT WHY?!?? right now I'm just blaming mysql
```

was written 13 years ago. My hope is that mysql solved this issue in all these years and we can remove this workaround that makes mysql slower in treebeard.

My concern is that we don't know when it was removed, and we may break somebody's install because they're running some really old mysql server version, even if we see tests pass in modern installs.

solarissmoke:
Currently, two separate queries are performed in MySQL when moving a node: the depth of nodes is updated separately. The reason is that MySQL handles assignments differently to standard SQL, with assignments processed left to right, using updated values for subsequent assignments.

From the MySQL [documentation](https://dev.mysql.com/doc/refman/8.4/en/update.html):

> If you access a column from the table to be updated in an expression, [UPDATE](https://dev.mysql.com/doc/refman/8.4/en/update.html) uses the current value of the column... This behavior differs from standard SQL. 

Rather than doing multiple queries, we can just adjust the update query to cater for this behaviour.

Fixes #199.

### DIFF
--- a/treebeard/mp_tree.py
+++ b/treebeard/mp_tree.py
@@ -250,29 +250,29 @@ class MP_ComplexAddMoveHandler(MP_AddHandler):
         """
 
         vendor = self.node_cls.get_database_vendor("write")
-        sql1 = "UPDATE %s SET" % (connection.ops.quote_name(get_result_class(self.node_cls)._meta.db_table),)
+        quoted_db_name = connection.ops.quote_name(get_result_class(self.node_cls)._meta.db_table)
 
-        if vendor == "mysql":
-            # hooray for mysql ignoring standards in their default
-            # configuration!
-            # to make || work as it should, enable ansi mode
-            # http://dev.mysql.com/doc/refman/5.0/en/ansi-mode.html
-            sqlpath = "CONCAT(%s, SUBSTR(path, %s))"
-        else:
-            sqlpath = sql_concat("%s", sql_substr("path", "%s", vendor=vendor), vendor=vendor)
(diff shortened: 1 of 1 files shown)


treebeard/mp_tree.py
```diff
@@ -250,29 +250,24 @@ def get_sql_newpath_in_branches(self, oldpath, newpath):
...
-            sql2.append(("depth=" + sql_length("%s", vendor=vendor) + "/%%s") % (sqlpath,))
-            vals.extend([newpath, len(oldpath) + 1, self.node_cls.steplen])
-        sql3 = "WHERE path LIKE %s"
-        vals.extend([oldpath + "%"])
-        sql = "%s %s %s" % (sql1, ", ".join(sql2), sql3)
+        if len(oldpath) != len(newpath):
+            # MySQL processes multiple assigments left to right, using the updated value
+            # for any column that is referenced in a subsequent assignment. This behavior differs from standard SQL.
+            # See https://dev.mysql.com/doc/refman/8.4/en/update.html
+            if vendor == "mysql":
```
freespace: It might be helpful here to have a simple example to illustrate the problem and how the code here fixes it. It is not entirely clear _how_ the MySQL branch gets around the problem.

PR #341, opened 2026-01-22
Remove all use of raw SQL in favour of ORM queries

### LINKED ISSUE #196: Add support for multiple databases
If there are multiple databases configured in Django, proper quote_name usage of the models is not ensured. There is a [first attempt at a solution](#166), but it is incomplete and only works for MP trees. It also broke travis. Ideally, we need to cache the value of the quoted value for each model so we don't have to find it every time it is needed.

### LINKED ISSUE #232: Treebeard, TreeAdmin, and Clustered Databases (Oh my!)
I'm using a clustered database setup, so writes are performed against the "write" database, and reads are performed against one of many "read" databases. It can take some time for writes to propagate from the write to the read databases, so if we need to perform a read operation directly after a write operation, we read directly from the "write" database.

This works in practice.

We're running into issues with the TreeAdmin page, however.

What I'm finding is that Node move operations seem to randomly break the view, creating weird `path` values on Nodes. Is there a way to assign a `using` call to TreeAdmin so every single call uses the same database?

EG, this is what we use elsewhere in the code to work around this:

`Node.objects.using("default").filter(pk=self.pk).update(code_path=Node.get_code_path(self))` (default is a direct connection to the write database)

### LINKED ISSUE #299: Add total descendants to queryset in django ORM
Hey, 

Not totally sure if I'm correct or not, but I seem to be able to annotate the total number of descendants to a MP queryset by using:

```
root_nodes.annotate(
    descendants_count=Subquery(
        Post.objects.filter(
            path__startswith=OuterRef("path"), depth__gt=OuterRef("depth")
        )
        .order_by()
        .annotate(count=Func(F("id"), function="Count"))
        .values("count")
    )
)
```

Post being my MP_Node model and root_nodes being a MP_Queryset.
This might be useful for the get_descendants_group_count method of the MP_Node.

https://github.com/django-treebeard/django-treebeard/blob/fe35cd2d33bed11946f8c28ee34aa2aaca2de73d/treebeard/mp_tree.py#L915

### LINKED ISSUE #300: Moving more than one node into parent node
Hi,

I'm using the 4.7 version and my model is:

```py
class Folder(MP_Node):
    node_order_by = ['name']

    path = models.CharField(max_length=4096, unique=True)
    name = models.CharField(max_length=200)
```

This is my file tree:

```py
Root (0001)
├── Child1 (00010001)
│   ├── SubChild1 (000100010001)
│   ├── SubChild2 (000100010002)
│   └── SubChild3 (000100010003)
└── Child2 (00010002)
```

When I try to move `SubChild1` and `SubChild2` with `sorted-child` into `Root` with:
```py
root = Folder.objects.get(name='Root')
folders_to_move = Folder.objects.filter(name__in=['SubChild1', 'SubChild2'])
for folder in folders:
    folder.move(root, pos='sorted-child')
```

It leads to:
```py
psycopg2.errors.CheckViolation: new row for relation "folder" violates check constraint "folder_numchild_check"
DETAIL:  Failing row contains (..., 000100010001, 2, -1, SubChild1).
```

The `numchild` is set in `mp_tree.py` by:

```py
    def get_sql_update_numchild(self, path, incdec='inc'):
        """:returns: The sql needed the numchild value of a node"""
        sql = "UPDATE %s SET numchild=numchild%s1"\
              " WHERE path=%%s" % (
                  connection.ops.quote_name(
                      get_result_class(self.node_cls)._meta.db_table),
                  {'inc': '+', 'dec': '-'}[incdec])
        vals = [path]
        return sql, vals
```

I think it occurs when it tries to reorder root children, the issue disappears when we use `last-child` without `node_order_by`.
Maybe `get_sql_update_numchild` should not decrement the `numchild` when this one is already 0 ?

solarissmoke:
This PR replaces all use of raw SQL queries in Treebeard with ORM queries. This is preferable because:

- It avoids potential issues with SQL injection: I found at least one instance where in specific (admittedly rare) circumstances injection could be performed;
- It means we can remove all the vendor-specific workarounds and leave it to Django itself to handle differences between databases;
- In a few cases it results in more efficient querying;
- Delegates selection of the right database, and things like quoting database name, to Django, simplifying our code considerably.

Fixes #196, fixes #300, fixes #299, fixes #232.

### DIFF
--- a/treebeard/mp_tree.py
+++ b/treebeard/mp_tree.py
@@ -5,9 +5,9 @@ import functools
 from typing import Any
 
 from django.core import serializers
-from django.db import connection, models, transaction
-from django.db.models import F, Q, Value
-from django.db.models.functions import Concat, Greatest, Substr
+from django.db import models, transaction
+from django.db.models import F, OuterRef, Q, Subquery, Value
+from django.db.models.functions import Concat, Greatest, Length, Substr
 from django.utils.translation import gettext_noop as _
 
 from treebeard.exceptions import InvalidMoveToDescendant, NodeAlreadySaved, PathOverflow
@@ -15,32 +15,9 @@ from treebeard.models import Node, get_result_class_base
 from treebeard.numconv import NumConv
 
 
-# The following functions generate vendor-specific SQL functions
-def sql_concat(*args, **kwargs):
-    vendor = kwargs.pop("vendor", None)
-    if vendor == "mysql":
-        return "CONCAT({})".format(", ".join(args))
-    if vendor == "microsoft":
-        return " + ".join(args)
-    return "||".join(args)
-
-
-def sql_length(field, vendor=None):
-    if vendor == "microsoft":
-        return f"LEN({field})"
-    return f"LENGTH({field})"
-
-
-def sql_substr(field, pos, length=None, **kwargs):
-    vendor = kwargs.pop("vendor", None)
-    function = "SUBSTR({field}, {pos})"
-    if length:
-        function = "SUBSTR({field}, {pos}, {length})"
-    if vendor == "microsoft":
-        if not length:
-            length = f"LEN({field})"
-        function = "SUBSTRING({field}, {pos}, {length})"
-    return function.format(field=field, pos=pos, length=length)
+class CountSubquery(Subquery):
+    template = "(SELECT COUNT(1) FROM (%(subquery)s) _count)"
+    output_field = models.PositiveIntegerField()
 
 
 get_result_class = functools.partial(get_result_class_base, identifying_field="path")
@@ -116,25 +93,12 @@ class MP_NodeManager(models.Manager):
         return MP_NodeQuerySet(self.model).order_by("path")
 
 
-class MP_AddHandler:
-    def __init__(self):
-        self.stmts = []
+class MP_ComplexAddMoveHandler:
+    def increment_numchild(self, path):
+        get_result_class(self.node_cls).objects.filter(path=path).update(numchild=F("numchild") + 1)
 
-
-class MP_ComplexAddMoveHandler(MP_AddHandler):
-    def run_sql_stmts(self):
-        cursor = self.node_cls._get_database_cursor("write")
-        for sql, vals in self.stmts:
-            cursor.execute(sql, vals)
-
-    def get_sql_update_numchild(self, path, incdec="inc"):
-        """:returns: The sql needed the numchild value of a node"""
-        sql = "UPDATE %s SET numchild=numchild%s1 WHERE path=%%s" % (
-            connection.ops.quote_name(get_result_class(self.node_cls)._meta.db_table),
-            {"inc": "+", "dec": "-"}[incdec],
-        )
-        vals = [path]
-        return sql, vals
+    def decrement_numchild(self, path):
+        get_result_class(self.node_cls).objects.filter(path=path).update(numchild=F("numchild") - 1)
 
     def reorder_nodes_before_add_or_move(self, pos, newpos, newdepth, target, siblings, oldpath=None, movebranch=False):
         """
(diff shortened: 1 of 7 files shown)


treebeard/mp_tree.py
```diff
@@ -148,113 +112,99 @@ def reorder_nodes_before_add_or_move(self, pos, newpos, newdepth, target, siblin
...
-                # For other databases: compute the depth using the new path value
-                set_sql.append(f"depth={sql_length(sqlpath, vendor=vendor)}/%s")
-                vals.extend([newpath, len(oldpath) + 1, self.node_cls.steplen])
-        vals.append(f"{oldpath}%")
-        sql = f"UPDATE {quoted_db_name} SET {', '.join(set_sql)} WHERE path LIKE %s"
-        return sql, vals
+            update_kwargs["depth"] = Length(new_path_value) / self.node_cls.steplen
+        update_kwargs["path"] = new_path_value
 
+        self.node_cls.objects.filter(path__startswith=oldpath).update(**update_kwargs)
```
gasman: This needs to be
```python
get_result_class(self.node_cls).objects.filter(path__startswith=oldpath).update(**update_kwargs)
```
or else it will fail to update descendants that are not of the same MTI subclass as the node being moved.

A test for this (to be added to `TestInheritedModels`):

```python
    def test_move_with_children(self, inherited_model):
        base_model = inherited_model.__bases__[0]

        node1 = base_model.objects.get(desc="1")
        node21 = inherited_model.objects.get(desc="21")
        node21.move(node1, "first-child")

        node1.refresh_from_db()
        node21.refresh_from_db()
        assert [node.desc for node in node1.get_children()] == ["21"]
        assert [node.desc for node in node21.get_children()] == ["211", "212"]
```

solarissmoke: Thanks @gasman, updated accordingly.

treebeard/ns_tree.py
```diff
@@ -203,7 +180,7 @@ def add_child(self, **kwargs):
             return new_sibling
 
         # we're adding the first child of this node
-        sql, params = self.__class__._move_right(self.tree_id, self.rgt, False, 2)
```
freespace: Not sure if it matters, but if I am reading the code correctly, previously changes to the database isn't done until `cursor.execute()`, whereas in the new code `_move_right()` will update the database by the time it returns. i.e. we are now updating the database earlier and the rest of the code didn't seem to change to reflect this, e.g. does `newobj._cached_parent_obj = self` still do what we think now that the database has changed before it executes?

freespace (approved):
Looks good to me. Flagged one area for additional consideration.

treebeard/ns_tree.py
```diff
@@ -155,37 +153,16 @@ def add_root(cls, **kwargs):
...
-            % {
-                "table": connection.ops.quote_name(get_result_class(cls)._meta.db_table),
-                "parent_rgt": rgt,
-                "tree_id": tree_id,
-                "lftop": lftop,
-                "incdec": incdec,
-            }
+        lftop = "lft__gte" if lftmove else "lft__gt"
+        output_field = models.PositiveIntegerField()
+        cls.objects.filter(rgt__gte=rgt, tree_id=tree_id).update(
```
gasman: Looks like these helpers also need a `get_results_class` (so that they're not only updating descendants that match the parent's class).

gasman:
Other than the potential missing `get_results_class` calls noted above, all looks good to me.

treebeard/ns_tree.py
solarissmoke: Looks like the caller is handling this in all cases (we have tests to cover).

treebeard/mp_tree.py
```diff
@@ -829,71 +733,20 @@ def get_root_nodes(cls):
...
-            node = cls(**dict(zip(field_names, node_data[:-2])))
-            node.descendants_count = node_data[-1]
-            ret.append(node)
-        return ret
+        :returns:
+
+            A Queryset of node objects with an extra attribute: `descendants_count`.
+        """
+        qs = parent.get_children() if parent else cls.get_root_nodes()
+        subquery = cls.objects.filter(path__startswith=OuterRef("path"))
```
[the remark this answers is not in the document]
solarissmoke: Thanks - added and also introduced a test to cover.

I am not keen on the way we have to call the `get_result_class` wrapper all over the place. At some point will see if I can find a cleaner way to do this across the board.

PR #356, opened 2026-02-05
Optimize load bulk

jpuglielli:
# MP_Node.load_bulk() Performance Improvements

## Summary

Rewrites `MP_Node.load_bulk()` to use `bulk_create()` for descendant nodes, significantly improving performance for large tree imports while maintaining full API compatibility.

## Key Changes

- **First-level nodes** still use `add_root()` and `add_child()` for correctness with concurrent operations
- **Descendant nodes** are bulk-created with pre-computed paths
- **Foreign keys** are bulk-fetched to avoid N+1 queries
- **Children** are sorted according to `node_order_by` before path assignment
- **Batch size parameter** added for controlling `bulk_create()` batch size
- **BulkNodeData TypedDict** added for better type safety

## Test Plan

- [x] All existing `load_bulk` tests pass
- [x] New tests for `node_order_by` sorting behavior
- [x] Full test suite passes (1210 tests)


solarissmoke (commented):
Hey @JoshPugli thanks for taking the time to work on this - your changes look really good! I have left a few minor comments. 

It also looks like the tests are failing on MySQL, which is always the most temperamental backend and may need some coaxing to behave as we want.

solarissmoke:
Superseded by a simpler implementation in #397

Issue #44, opened 2016-11-23
Use a Model Manager for object creation


jrief:
Currently, method ``add_root`` is implemented as a class-method of the current model node. This is not Djangonic behavior. In Django, creating a new model instance, normally is done through the model manager and not through a class- or static method.

zbyte64:
Expanding on this, class methods that return a queryset like `get_root_nodes` & `get_ancestors` should also be implemented as a queryset.

tabo:
What do you suggest? How would this API look like?

PR #371, opened 2026-02-17
[WIP] Move table-level methods into managers

solarissmoke:
Move the following methods that are not specific to a single model instance into the manager:

- `add_root`
- `load_bulk`
- `dump_bulk`
- `get_root_nodes`
- `get_first_root_node`
- `get_last_root_node`
- `find_problems`
- `fix_tree`
- `get_tree`
- `get_descendants_group_count`

Having these methods on the manager, rather than as class methods on the model, is consistent with the approach that Django takes for models (e.g., `Model.objects.create()`), and aligns Treebeard more closely with that pattern.

Refs #44

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -8,30 +8,11 @@ API
   :show-inheritance:
 
   This is the base class that defines the API of all tree models in this
-  library:
-
-     - :class:`treebeard.mp_tree.MP_Node` (materialized path)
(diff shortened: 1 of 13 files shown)


solarissmoke:
Parking this for now. While I can see the appeal of aligning with Django model managers, the down side is that many projects don't currently replace the default manager with their own, and upgrading would mean they need to subclass the treebeard manager. I don't think it's a huge burden, but it is probably worth getting feedback from the community before implementing.

PR #387, opened 2026-05-08
Add a find_problems method for NS_Node

gasman:
Add a `find_problems` method to `NS_Node`, similar to the existing one for `MP_Node`. This catches badly-nested `lft`/`rgt` ranges, incorrect `depth` values, and multiple root nodes with the same `tree_id`. Note that gaps in the `lft`/`rgt` sequence are _not_ considered a problem here.


gasman:
Closing as this is being reviewed as part of #388.

PR #388, opened 2026-05-08
 Fix NS_Node.add_child and add_sibling on inherited models 

gasman:
The `add_child` and `add_sibling` methods on `NS_Node` did not properly account for model inheritance, as the queries for updating `lft`/`rgt`/`tree_id` across the tree were being run against the class of the initial instance, rather than the base model. As a result, `some_node.add_child(...)` will fail to update those fields on nodes that do not match the specific class of `some_node`.

Incorporates #387 for testing.

### DIFF
--- a/tests/test_treebeard.py
+++ b/tests/test_treebeard.py
@@ -129,6 +129,11 @@ def mpsmallstep_model(request):
     return request.param
 
 
+@pytest.fixture(scope="function", params=[models.NS_TestNode])
+def ns_model(request):
+    return request.param
+
+
 class TestTreeBase:
     def got(self, model):
         if model in [models.NS_TestNode, models.NS_TestNode_Proxy]:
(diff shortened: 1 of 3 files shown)


treebeard/ns_tree.py
```diff
@@ -561,6 +565,82 @@ def get_root_nodes(cls):
...
+            qs = cls.objects.all()
+            initial_depth = 0
+
+        reversed_lft_rgt, overlapping_nodes, duplicate_tree_ids, wrong_depth = [], [], [], []
+
+        # Iterate over the nodes in order of tree_id and lft
+        qs = qs.order_by("tree_id", "lft").values("pk", "tree_id", "lft", "rgt", "depth")
+
+        # Consider each tree individually
+        for tree_id, nodes in groupby(qs, lambda x: x["tree_id"]):
```
solarissmoke: nit: `operator.itemgetter("tree_id")` is a little more efficient than a lambda function.

treebeard/ns_tree.py
```diff
@@ -561,6 +565,82 @@ def get_root_nodes(cls):
...
+            qs = cls.objects.filter(tree_id=parent.tree_id, lft__gte=parent.lft, rgt__lte=parent.rgt)
+            initial_depth = parent.get_ancestors().count()
+        else:
+            qs = cls.objects.all()
+            initial_depth = 0
+
+        reversed_lft_rgt, overlapping_nodes, duplicate_tree_ids, wrong_depth = [], [], [], []
+
+        # Iterate over the nodes in order of tree_id and lft
+        qs = qs.order_by("tree_id", "lft").values("pk", "tree_id", "lft", "rgt", "depth")
```
solarissmoke: Do we want an `iterator()` here? I know we load everything into memory below anyway, but this would still reduce memory footprint.

tests/test_treebeard.py
```diff
@@ -2118,6 +2126,51 @@ def test_add_root_with_node_order_by(self, inherited_model_with_sort):
...
+        # We need a leaf node of inherited_model to test against, as non-leaves will delegate to add_sibling instead
+        node21 = inherited_model.objects.get(desc="21")
+        node213 = inherited_model(desc="213")
+        node21.add_child(instance=node213)
+
+        node213.add_child(desc="2131")
+
+        assert [node.desc for node in node213.get_children()] == ["2131"]
+
+        if issubclass(inherited_model, (MP_Node, NS_Node)):
```
solarissmoke: tempted to have a helper function for this so that if/when `find_problems()` is added to another class, we only have to update it one place.

solarissmoke (commented):
Thanks @gasman, looks good! A couple of minor comments. Have reviewed `find_problems()` in this PR as well.

treebeard/ns_tree.py
gasman: Updated in 7883e92915d6fec95d2e886e6c983c1fced9367f.

treebeard/ns_tree.py
gasman: TIL! Updated in 7883e92915d6fec95d2e886e6c983c1fced9367f.

(Do we load the full queryset into memory further down? I can't see where...)

tests/test_treebeard.py
gasman: Good plan - added an `_assert_no_tree_problems` method in e3cbbc134a3f54134a7393631bcf6c9a3aa8fca8.

treebeard/ns_tree.py
solarissmoke: No, sorry, I was wrong about loading everything into memory further down - failed to recognise that groupby itself returns an iterator.

solarissmoke (approved):
Thanks! Will cut a bugfix release tomorrow.

solarissmoke:
Released in v5.1.0

PR #390, opened 2026-05-16
Fix node paths being incremented during noop move operations

### LINKED ISSUE #389: Path updating every time a record is saved in the admin / duplicate key error on add new
I revealed my paths in the admin to better understand what was going on, and I've noticed that every time I save a record, the path is incremented. After 10+ saves, the same record has gone from a path of 0005 to 000K. I have no idea why this would be happening. Any thoughts? Seems like this would quickly whittle down available paths the longer this goes on, leaving substantial gaps between certain records.

solarissmoke:
PR has two commits:

- First fixes handling of noop move operations for MP_Nodes with node_order_by so that the paths of these nodes are not incremented when the node doesn't require moving. Also cleans up logic for short-circuit during move to make it easier to understand. Some comments inline.

- Second modifies the form used in the admin to just skip all treebeard-specific logic entirely unless the fields treebeard cares about have changed. The queries for moving nodes are expensive and we shouldn't run them if we know nothing has changed.

Fixes #389

### DIFF
--- a/tests/test_treebeard.py
+++ b/tests/test_treebeard.py
@@ -129,6 +129,11 @@ def mpsmallstep_model(request):
     return request.param
 
 
+@pytest.fixture(scope="function", params=[models.MP_TestSortedNodeShortPath])
+def mpsorted_model(request):
+    return request.param
+
+
 @pytest.fixture(scope="function", params=[models.NS_TestNode])
 def ns_model(request):
     return request.param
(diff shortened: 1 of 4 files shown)


tests/test_treebeard.py
```diff
@@ -2307,19 +2350,6 @@ def test_move(self, mpsmallstep_model):
                     newroot.move(target, pos)
 
 
-@pytest.mark.django_db
-class TestMP_TreeShortPath(TestTreeBase):
-    """Test a tree with a very small path field (max_length=4) and a
-    steplen of 1
-    """
-
-    def test_short_path(self, mpshortnotsorted_model):
```
solarissmoke, line 2316: Just moving into a shared test class above.

treebeard/mp_tree.py
```diff
@@ -361,24 +359,35 @@ def process(self):
         if self.target.is_descendant_of(self.node):
             raise InvalidMoveToDescendant(_("Can't move node to a descendant."))
 
-        if oldpath == self.target.path and (
```
solarissmoke, line 364: This logic has been moved further down (see "Handle special cases where nothing needs to be done"), and tweaked slightly. Specifically, it previously only ran for the _very niche_ edge case of moving a node relative to itself (`oldpath == self.target.path` can only be the case if `self.node` and `self.target` are identical).

I believe this can be broadened to handle a broader set of noop cases:

- Moving a node to `first-sibling` position, when it is already the first sibling of the target;
- Moving a node to `last-sibling` position, when it is already the last sibling of the target.

I am relying on existing test coverage here, which I think is reasonably complete.

treebeard/mp_tree.py
```diff
@@ -361,24 +359,35 @@ def process(self):
...
-            return
-
         if self.pos == "sorted-sibling":
             siblings = self.node.get_sorted_pos_queryset(self.target.get_siblings(), self.node)
-            first = siblings.first()
-            newpos = first._get_lastpos_in_path() if first else None
-            if newpos is None:
+            if first := siblings.first():
+                newpos = first._get_lastpos_in_path()
+                if self.node._get_lastpos_in_path() == newpos - 1:
```
solarissmoke, line 366: This detects the case where you are asking to move a node sorted by some user-defined field, and the sorting would place it in the same position that it is already in. We have to short-circuit, otherwise the path would get incremented.

treebeard/mp_tree.py
```diff
@@ -351,8 +351,6 @@ def __init__(self, node, target, pos=None):
     def process(self):
         self.pos = self.node._prepare_pos_var_for_move(self.pos)
 
-        oldpath = self.node.path
```
solarissmoke, line 354: This was unnecessary AFAICT, since we aren't modifying `self.node` anywhere until we execute the final query.

tests/test_treebeard.py
```diff
@@ -2687,6 +2717,27 @@ def test_form_admin(self, model):
             nodes = self._get_nodes_list(safe_parent_nodes)
             self._assert_nodes_in_choices(form, nodes)
 
+    def test_skips_move_if_no_change(self, model_without_data):
```
freespace, line 2720: Can probably also test the path involving `node_ordered_by`

freespace (approved):
The logic for skipping computation when form changes do not impact treebeard look correct to me. It would bit a little extra effort to cover the cases where the changing the sorted-by fields should also trigger recomputation.

PR #391, opened 2026-05-19
feat(mp_tree): add path_updated signal for bulk path updates

### LINKED ISSUE #381: Implement a signal that fires when path is bulk-updated
In Wagtail (or specifically the [django-modelsearch](https://github.com/wagtail/django-modelsearch) package spun out of it), we support Elasticsearch and OpenSearch as search backends by maintaining an external data store that mirrors the data in the database. This is kept in sync with the database through `post_save`/`post_delete` signals on the models being tracked.

To allow this data store to be queried using the same kinds of tree operations that treebeard allows on the database - for example, searching pages within a particular subtree - we would like to include the `path` and `depth` fields in the data stored there. However, since treebeard uses `queryset.update` to bulk-update these fields in move operations, we cannot rely on the `post_save`/`post_delete` signals to keep these in sync with the database.

To address this, I propose that treebeard should implement a `path_updated` signal that is fired whenever treebeard performs a `queryset.update` operation that replaces some prefix of `path` with a new prefix. This would pass the old and new path prefixes as arguments. `django-modelsearch` could then listen to this signal along with `post_save` and `post_delete` for any model subclassing `MP_Node`, and perform the equivalent update on the Elasticsearch/OpenSearch data store.

A similar signal could be implemented for `NS_Node` when updating the `lft`/`rgt` values.

Does this sound like something django-treebeard would be willing to include?

SAY-5:
Closes #381. Adds `treebeard.mp_tree.path_updated`, sent whenever an `MP_Node` operation rewrites a path prefix on a subtree via a bulk `UPDATE` (the queries inside `move`, sorted `add_sibling`, and `fix_tree`). Receivers get `sender`, `old_path` and `new_path` so external stores can mirror the rewrite without subscribing to every `post_save`.

The signal is defined next to `mp_tree.py` because the prefix-replacement shape is specific to materialized paths. `NS_Node` does the equivalent work by rewriting individual `lft`/`rgt`/`tree_id` values rather than replacing a prefix, so there is no single `(old, new)` pair to broadcast; the LTree backend mutates each affected path independently (appending an `A` label) for the same reason; `AL_Node` has no path concept at all. Happy to add tree-id or lft/rgt level signals for `NS_Node` in a follow-up if that shape is useful.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@ Next release (in development)
 
 * Fixed `MP_Node` paths being incremented unnecessarily when saving nodes with `node_order_by` in the admin.
 * Optimised `MoveNodeForm` to skip move logic if no treebeard-specific fields were modified.
+* Added `treebeard.mp_tree.path_updated` signal, sent whenever an `MP_Node` bulk update rewrites a path prefix on a subtree.
 
 
 Release 5.1.0 (May 12, 2026)
--- a/docs/source/mp_tree.rst
+++ b/docs/source/mp_tree.rst
@@ -267,6 +267,35 @@ extra steps, materialized path is more efficient than other approaches.
   :show-inheritance:
 
 
+Signals
+-------
+
+.. data:: path_updated
+
+   Sent whenever an :class:`MP_Node` operation rewrites a path prefix on a
+   subtree via a bulk ``UPDATE`` (the queries used inside ``move``,
+   sorted ``add_sibling`` and ``fix_tree``). These updates bypass
+   ``post_save``, so applications that mirror tree data into an external
+   store (search index, cache, etc.) can listen for ``path_updated`` to
+   keep that store in sync.
+
+   Provides ``sender`` (the tree model class), ``old_path`` (the path
+   prefix that was replaced) and ``new_path`` (the prefix it was
+   replaced with). Every row whose ``path`` started with ``old_path`` now
+   starts with ``new_path``.
+
+   Example:
+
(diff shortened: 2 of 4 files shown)


SAY-5:
Fair points, all taken. I have force-pushed a rewrite that addresses each of them.

The signal now lives in `treebeard.mp_tree` next to the code that sends it, and the two send sites are unconditional (no guards on `old_path`/`new_path`, since by the time we reach the bulk `UPDATE` there is a real prefix shift to report). `test_no_signal_without_receiver` is gone. The tests now use a receiver that captures `(sender, old_path, new_path)` for every emission and asserts against exact expected tuples, with `assert len(...) == N` before any indexing so the "iterate-on-possibly-empty-list" trap is closed. They cover `move(last-sibling)`, `add_sibling` (`left`, `first-sibling`, `last-sibling`), sorted `add_sibling` and `move` on `MP_TestSortedNodeShortPath` (which uses `node_order_by`), `fix_tree`, and a no-op move that must not emit.

On NS/LTree: `NS_Node` rewrites individual `lft`/`rgt`/`tree_id` values rather than substituting a prefix, so there is no natural `(old, new)` pair to broadcast on the same signal; the LTree backend mutates each affected path independently (appending an `A` label per row) for the same reason. I noted this in the PR body and am happy to add an `NS_Node` shift signal in a follow-up if it would be useful; the design there is a separate conversation. If anything still misses the mark, please point at the line and I will fix it.

PR #393, opened 2026-05-21
ltree: Do not update non-sibling paths in _move_subtree_right

gasman:
The `_move_subtree_right` method (which increments the path labels of nodes so that a sibling can be inserted before them where no gap exists in the lexicographic ordering) failed to limit the update to siblings of the new node; as a result, nodes at the same depth in entirely different subtrees were being redundantly updated.


solarissmoke (approved):
Thanks! Nice to see that somebody is using the ltree implementation.

PR #395, opened 2026-05-22
Simplify logic for NS delete

gasman:
* `merge_deleted_counters` and the `deleted_counter` argument are redundant - deletion only ever happens in a single query, so the only thing it ever merges with is the empty `(0, {})` value
* When eliminating descendants from the queryset, we know that the only possible ancestor is the last node added to the list, so there's no need to scan the whole list
* The descendant check can be done at the same time as building up `toremove` and `ranges`, and this can be done on a values queryset to avoid instantiating nodes in full.

### DIFF
--- a/treebeard/ns_tree.py
+++ b/treebeard/ns_tree.py
@@ -13,14 +13,6 @@ from treebeard.exceptions import InvalidMoveToDescendant, NodeAlreadySaved
 from treebeard.models import Node
 
 
-def merge_deleted_counters(c1, c2):
-    """
-    Merge return values from Django's Queryset.delete() method.
-    """
-    object_counts = {key: c1[1].get(key, 0) + c2[1].get(key, 0) for key in set(c1[1]) | set(c2[1])}
-    return (c1[0] + c2[0], object_counts)
-
-
 class NS_NodeQuerySet(models.query.QuerySet):
     """
     Custom queryset for the tree node manager.
@@ -28,7 +20,7 @@ class NS_NodeQuerySet(models.query.QuerySet):
     Needed only for the customized delete method.
     """
 
-    def delete(self, *args, removed_ranges=None, deleted_counter=None, **kwargs):
+    def delete(self, *args, **kwargs):
         """
         Custom delete method, will remove all descendant nodes to ensure a
         consistent tree (no orphans)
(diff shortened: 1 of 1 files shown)


treebeard/ns_tree.py
```diff
@@ -56,32 +44,30 @@ def delete(self, *args, removed_ranges=None, deleted_counter=None, **kwargs):
...
+                # Remove this node and its descendants
+                toremove.append(Q(lft__range=(node["lft"], node["rgt"])) & Q(tree_id=node["tree_id"]))
+                ranges.append((node["tree_id"], node["lft"], node["rgt"]))
+
             if toremove:
-                deleted_counter = model.objects.filter(reduce(operator.or_, toremove)).delete(
-                    removed_ranges=ranges, deleted_counter=deleted_counter
-                )
-        return deleted_counter
+                return model.objects.filter(reduce(operator.or_, toremove)).delete(removed_ranges=ranges)
```
solarissmoke: This can be simplified even further to remove the recursive call to `delete()` and `removed_ranges` kwarg entirely:

```python
result = super(NS_NodeQuerySet, model.objects.filter(reduce(operator.or_, toremove))).delete()
# Move logic above for closing holes into here:
for tree_id, drop_lft, drop_rgt in sorted(ranges, reverse=True):
    model._close_gap(drop_lft, drop_rgt, tree_id)
return result
```

treebeard/ns_tree.py
```diff
@@ -56,32 +44,30 @@ def delete(self, *args, removed_ranges=None, deleted_counter=None, **kwargs):
...
+            else:
+                return (0, {})
```
solarissmoke: ~nit: prefer to delegate return value construction to Django with:~

```python
return model.objects.none().delete()
```

Edit: ignore me - this would cause infinite recursion unless it was delegated to the super class. Might be worth moving to a guard clause at the top, that returns early, instead of right at the bottom.

solarissmoke (changes requested):

treebeard/ns_tree.py
gasman: Good call - updated in 83607cf9cbf4b190e4f30e59a29088a05f73f2ae. I considered this earlier, but wasn't totally confident of getting the details of the `super` call right, given that we're potentially switching the queryset to a different model in the call to `tree_model()` - however, I now see that `MP_Node` is already doing the same thing.

PR #396, opened 2026-05-25
Add signals for bulk updates on MP_Node, NS_Node, LT_Node

### LINKED ISSUE #381: Implement a signal that fires when path is bulk-updated

gasman:
Fixes #381; built on top of #395.

Implement the following signals that are fired after performing bulk updates or deletions on a tree:

* `MP_Node`: `path_updated`, `nodes_deleted`
* `NS_Node`: `gap_altered`, `tree_ids_incremented`, `subtree_moved`, `nodes_deleted`
* `LT_Node`: `subtree_moved_right`, `subtree_moved`, `nodes_deleted`

These should be sufficient to keep an external data store such as a search index in sync with the database.

### DIFF
--- a/docs/source/ltree.rst
+++ b/docs/source/ltree.rst
@@ -112,5 +112,71 @@ To use the ``ltree`` module, you need to create the extension in your database:
 .. autoclass:: LT_NodeQuerySet
   :show-inheritance:
 
+Signals
+-------
+
+The :mod:`treebeard.ltree` module defines several signals that are sent when
+bulk updates are made to the tree. Along with the standard Django ``post_save``
+and ``post_delete`` signals that track changes to individual node instances,
+these can be used to keep external data stores such as search indexes in sync
+with the tree.
+
+.. attribute:: subtree_moved_right
+
+   Sent after a bulk update has been performed to increment existing path
+   values to allow inserting a sibling node, with the following arguments:
+
+   ``sender``
+      The model class where the update occurred.
+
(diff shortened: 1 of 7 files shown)


solarissmoke:
Thanks! Will ship a release with the recent changes next week

PR #398, opened 2026-05-28
Make uniqueness constraint on LTree.path deferrable

solarissmoke:
Makes the uniqueness constraint for ltree paths deferrable, so that operations like `_move_subtree_right` don't fail immediately. As per postgresql [documentation](https://www.postgresql.org/docs/current/sql-createtable.html#id-1.9.3.85.9.4):

> When a UNIQUE or PRIMARY KEY constraint is not deferrable, PostgreSQL checks for uniqueness immediately whenever a row is inserted or modified. The SQL standard says that uniqueness should be enforced only at the end of the statement; this makes a difference when, for example, a single command updates multiple key values. To obtain standard-compliant behavior, declare the constraint as DEFERRABLE but not deferred (i.e., INITIALLY IMMEDIATE). Be aware that this can be significantly slower than immediate uniqueness checking.

This does come at a performance cost, but it seems like it would be costlier to have to work around this manually.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@ Next release (in development)
 
 * Fixed `MP_Node` paths being incremented unnecessarily when saving nodes with `node_order_by` in the admin.
 * Optimised `MoveNodeForm` to skip move logic if no treebeard-specific fields were modified.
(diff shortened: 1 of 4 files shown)


solarissmoke:
@gasman flagging this for your awareness, as you seem to be using the ltree implementation!

gasman:
@solarissmoke Thanks! I'm not actively using ltree though - I just dipped into it for the signals implementation :-)

PR #400, opened 2026-06-02
Add checks to ensure manager methods subclass Treebeard managers

solarissmoke:
This is documented, but not currently enforced, and blocks us from implementing #44 (moving class methods currently on the model class to the manager, for consistency with how Django itself handles row-level methods).

Some downstream projects, notably Wagtail, do not currently subclass Treebeard's manager, making it impossible to move these methods in a backwards-compatible way. This it seems prudent to add a warning first.

### DIFF
--- a/tests/test_treebeard.py
+++ b/tests/test_treebeard.py
@@ -7,12 +7,15 @@ from unittest import mock
 from unittest.mock import patch
 
 import pytest
+from django.apps import apps
 from django.contrib.admin.options import TO_FIELD_VAR
 from django.contrib.admin.sites import AdminSite
 from django.contrib.admin.views.main import ChangeList
 from django.contrib.auth.models import AnonymousUser, User
 from django.contrib.messages.storage.fallback import FallbackStorage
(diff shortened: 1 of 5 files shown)


solarissmoke:
@gasman would be grateful for your eyes on this change, as it will impact Wagtail (which doesn't currently subclass `MP_NodeManager` in its tree managers - I will make a PR to do that). 

The rationale is described in #44: Treebeard classes currently expose a bunch of class methods (e.g., `get_tree`, `get_root_nodes`, `get_first_root_node`) that should really be methods on the model manager for consistency with Django's design patterns. Having them on the manager also allows much more fine-grained control over their use - e.g., selecting which database to use, which is not currently possible.

The plan is to move those methods to the manager, but retain backwards compatibility for some time by retaining the existing class methods that delegate to the manager. Of course, this only works if the manager subclasses the Treebeard manager that holds the new methods.

It is documented [here](https://django-treebeard.readthedocs.io/en/latest/caveats.html#overriding-the-default-manager) that the treebeard manager should always be subclassed. This change makes it more explicit.

PR #406, opened 2026-06-18
fix: clear prefetch_related before iterator() in tree delete methods (fixes #405)

### LINKED ISSUE #405: MP Tree delete fails when queryset has `prefetch_related`
[This line](https://github.com/django-treebeard/django-treebeard/blob/3d0c1fc390defdec40b14eb648e5dc8659711735/treebeard/mp_tree.py#L42) raises a `ValueError` with the message:

> chunk_size must be provided when using QuerySet.iterator() after prefetch_related().

The cause is an added `prefetch_related` to `get_queryset` of the tree admin to avoid N+1 queries like so:

```python
@admin.register(Segment)
class SegmentAdmin(TreeAdmin):
    # ...
    def get_queryset(self, request):
        qs = super().get_queryset(request)
        qs = qs.prefetch_related("tags")
        return qs
```

Ideally, I would keep this improvement and django-treebeard could provide the `chunk_size` to `.iterator()` (as the exception suggests) or clear the `prefetch_related` which is likely unnecessary for the delete operation. There are also additional code paths that use `.iterator()` which may also be affected.

gaoflow:
When a user adds `prefetch_related` to a tree admin `get_queryset` (e.g. to avoid N+1 queries in the admin list view), the queryset `delete()` crashed because `.iterator()` refuses to iterate when prefetch lookups are pending:

```
ValueError: chunk_size must be provided when using
QuerySet.iterator() after prefetch_related()
```

The three tree queryset classes (`MP_NodeQuerySet`, `NS_NodeQuerySet`, `LT_NodeQuerySet`) all call `.iterator()` in their custom `delete()` to enumerate nodes for removal. These delete paths only access `path`/`depth`/`numchild` (via `.only()`) or `tree_id`/`lft`/`rgt` (via `.values()`) — prefetched related objects are never used during deletion.

Calling `.prefetch_related(None)` before `.iterator()` clears any user-added prefetch lookups, which is safe because the delete path does not use related-object access.

Fixes #405.

### DIFF
--- a/treebeard/ltree/__init__.py
+++ b/treebeard/ltree/__init__.py
@@ -106,7 +106,7 @@ class LT_NodeQuerySet(models.query.QuerySet):
         """
         # Construct the minimal list of nodes that need deleting along with their descendants
         paths_to_remove = set()
-        for node in self.order_by("path").only("path").iterator():
+        for node in self.order_by("path").only("path").prefetch_related(None).iterator():
             found = False
             for depth in range(1, len(node.path)):
                 path = node.path[0:depth]
--- a/treebeard/mp_tree.py
+++ b/treebeard/mp_tree.py
@@ -39,7 +39,7 @@ class MP_NodeQuerySet(models.query.QuerySet):
         # to be deleted and remove nodes from the list if an ancestor is
         # already getting removed, since that would be redundant
         removed = {}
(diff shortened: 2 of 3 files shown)


solarissmoke:
Thanks for this! I do want to add regression tests, but in attempted to do so discovered an unrelated issue, so will do that separately.

PR #408, opened 2026-06-24
Move table-level methods into managers

### LINKED ISSUE #44: Use a Model Manager for object creation

solarissmoke:
Moves table-level functionality that was previously implemented as methods on the model class, to methods on the model manager, which is the [preferred pattern](https://docs.djangoproject.com/en/stable/topics/db/managers/#adding-extra-manager-methods) with Django. Retains backward-compatibility on the old methods, which will be dropped in the next major release.

This change will be shipped as a major release.

The backward compatibility only works if a project using Treebeard has not replaced the default model manager with one that doesn't subclass Treebeard's manager. A [warning was added for this case](https://github.com/django-treebeard/django-treebeard/commit/bd8b2483c359f0ad459b0c36f21e5a5b6e0036b1) in the previous release, which here is changed to an error. We are relying on such projects pinning at least major versions for this not to break things.

I have duplicated the old test suite to use the deprecated API, to make sure back-compat works.

Most of the diff here is just functions moving from one place to another. I have noted in comments the areas that I think would benefit from review.

Fixes #44

### DIFF
--- a/treebeard/models.py
+++ b/treebeard/models.py
@@ -3,11 +3,14 @@
 import operator
 import warnings
 from contextlib import suppress
-from functools import cache, reduce
+from functools import reduce
 
+from django.core import checks
 from django.db import models, transaction
 from django.db.models import Q
+from django.utils.functional import cached_property
 
+from treebeard.deprecation import RemovedInTreebeard7Warning, get_moved_manager_classmethod, get_moved_manager_method
 from treebeard.exceptions import InvalidPosition, MissingNodeOrderBy
 from treebeard.utils import prepare_dumpdata_for_loading, save_m2m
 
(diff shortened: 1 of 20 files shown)


treebeard/models.py
```diff
@@ -22,17 +25,201 @@ class Node(models.Model):
...
+        return errors
+
+    class Meta:
+        """Abstract model."""
+
+        abstract = True
+
+
+# Deprecated class methods that have moved to the model manager. Will be removed in Treebeard 7
+def _inject_moved_method_back_compat():
```
solarissmoke, line 137: Shim to inject backward compatible methods into the model class. Sanity check on this and the associated code in `treebeard/deprecation.py` would be appreciated.

treebeard/utils.py
```diff
@@ -36,3 +39,32 @@ def save_m2m(node: models.Model, deserialized_obj: DeserializedObject):
         for accessor_name, object_list in deserialized_obj.m2m_data.items():
             getattr(node, accessor_name).set(object_list)
         deserialized_obj.m2m_data = None  # Avoid accidental reuse of m2m_data if save() is called on the object
+
+
+def check_create_args(func):
```
solarissmoke, line 44: DRYed out checks that are needed for all methods that create objects into a single decorator.

All create methods accept data for new objects in two ways: either a dictionary of field values, or an unsaved model instance constructed with `MyModel(**kwargs)`. 

This decorator:

- Ensures that exactly one of these is supplied;
- If `instance` is supplied, ensures that it is unsaved and doesn't already have a row in the database.

Would appreciate sanity check.

freespace: As far as I can tell the code does what it says on the tin.

treebeard/models.py
freespace: Code checks out after reminding myself how `partial()` works xD

Nitpick would be that the methods `moved_manager_classmethod` and `moved_manager_method` could benefit from a `get_` verb-prefix to bring it into stylistic consistency with the surrounding code, e.g. `_inject`, `_handle`.

freespace (approved):

PR #415, opened 2026-08-18
Fix silent no-op on sorted move across parents

### LINKED ISSUE #414: Sorted move silently no-ops when the node's last path segment matches the target slot
## Summary

On a tree with `node_order_by`, `MP_NodeManager.move(node, target, pos='sorted-child')` (and `sorted-sibling`) can **silently do nothing**: no exception, no move. 
It happens whenever the moved node's *last path segment* happens to equal `newpos - 1` of the computed target slot, even though the node lives under a **completely different parent**.

Because `MoveNodeForm` / `TreeAdmin` use this code path, the admin reports a successful save while the node never moves, with no traceback anywhere, which makes this quite confusing for end users.

## Version

- django-treebeard 7.0.0
- Django 6.0.x
- Python 3.14
- SQLite (backend-independent, the faulty check is pure path arithmetic)

## Reproduction

```python
class Category(MP_Node):
    name = models.CharField(max_length=50)
    node_order_by = ['name']
```

```python
root = Category.add_root(name='root')                            # 0001
parent = Category.objects.get(name='root').add_child(name='parent')  # 00010001
for name in ('c1', 'c2', 'c3'):
    Category.objects.get(name='parent').add_child(name=name)
# c1/c2/c3 = 000100010001 / ...0002 / ...0003

# Move c1 up under root — works ('c1' sorts before 'parent'):
c1, root = Category.objects.get(name='c1'), Category.objects.get(name='root')
Category.objects.move(c1, root, pos='sorted-child')
# c1 is now 00010001, parent shifted to 00010002

# Move c1 back under parent — SILENTLY DOES NOTHING:
c1 = Category.objects.get(name='c1')
parent = Category.objects.get(name='parent')
Category.objects.move(c1, parent, pos='sorted-child')

c1 = Category.objects.get(name='c1')
assert c1.get_parent().name == 'parent'   # FAILS — c1 is still a child of root
```

The same tree/moves through the admin (`MoveNodeForm`) show the same behaviour: the change form saves with a success message, but the node stays where it was.

## Root cause

`MP_NodeManager.move()` has a "nothing to do" shortcut in the `sorted-sibling` branch (which `sorted-child` is converted into by `update_move_to_child_vars`), present in 7.0.0 and still on current master:
[treebeard/mp_tree.py#L499-L505](https://github.com/django-treebeard/django-treebeard/blob/8422425b28b8651e52fdae5a9c2bb91bf80709bd/treebeard/mp_tree.py#L499-L505)

```python
if pos == "sorted-sibling":
    siblings = self.get_sorted_pos_queryset(self.get_siblings(target), node)
    if first := siblings.first():
        newpos = first._get_lastpos_in_path()
        if node._get_lastpos_in_path() == newpos - 1:
            # The node is already in the right place, nothing to do
            return
```

The check compares **only the last path segment** of the node against the target slot.
It never verifies that the node is actually a sibling of `first` (same parent path prefix).
In the repro: `c1` sits at `0001·0001` (lastpos 1) and the slot before `c2` (`…·0002`, lastpos 2) computes `newpos - 1 == 1` 
→ treated as "already in the right place" although the node is under a different parent.

The bug is selective, a node whose lastpos does not collide (e.g. `c3` in the same tree) moves back perfectly, which makes the behaviour look random from the admin.

## Related issues

Not a duplicate of [#300](https://github.com/django-treebeard/django-treebeard/issues/300) (numchild constraint violation on sorted moves, fixed in [#341](https://github.com/django-treebeard/django-treebeard/pull/341)) or [#145](https://github.com/django-treebeard/django-treebeard/issues/145) (duplicate key on sorted-child). 
Those fail loudly, this one silently skips the move.

domibarton:
The "already in the right place" shortcut in MP_NodeManager.move() only compared the node's last path segment against the computed target slot. When the segment matched (e.g. moving a node back under its previous parent), the move was silently skipped even though the node lived under a completely different parent.

Also require the node to share the parent path with the computed sibling before treating the move as a no-op.

Fixes #414

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -17,7 +17,7 @@ Contributions made by:
     * Alexei Vlasov
     * moberley
     * czare1
-    * Fernando Gutierrez (xbito) 
+    * Fernando Gutierrez (xbito)
     * Jens Diemer
     * Jacob Rief
     * Maik Hoepfel
@@ -25,3 +25,4 @@ Contributions made by:
     * Johannes Wilm
     * Awais Qureshi
     * Julian Wachholz
+    * Dominique Barton
--- a/tests/test_treebeard.py
+++ b/tests/test_treebeard.py
@@ -3254,6 +3254,39 @@ class TestTreeSorted(TestTreeBase):
         ]
         assert self.got(sorted_model) == expected
 
+    def test_move_sorted_back_under_previous_parent(self, sorted_model):
+        # Regression test: a sorted move used to be silently skipped when the
+        # node's last path segment happened to match the computed target slot,
+        # even though the node lived under a different parent.
+        # For more information see:
+        # https://github.com/django-treebeard/django-treebeard/issues/414
+        root = sorted_model.objects.add_root({"val1": 1, "val2": 1, "desc": "rt"})
+        parent = sorted_model.objects.add_child(root, {"val1": 2, "val2": 1, "desc": "prt"})
+        for val1, desc in ((1, "ac1"), (2, "ac2"), (3, "ac3")):
+            parent = sorted_model.objects.get(pk=parent.pk)
+            sorted_model.objects.add_child(parent, {"val1": val1, "val2": 1, "desc": desc})
+
+        # Move ac1 up under the root: it sorts before its old parent, taking
+        # the position whose last path segment collides with the slot computed
+        # for the move back below.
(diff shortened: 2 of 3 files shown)


solarissmoke:
Thanks for the report and fix - this was a sneaky one indeed! Will ship a bugfix shortly.

PR #417, opened 2026-08-28
Benchmarking Ancestors

maklem:
Hi,

I have a use case where getting ancestors is the most common operation. While trying to add a benchmark I noticed `tests/test_benchmarks.py` was [recently migrated to use managers](https://github.com/django-treebeard/django-treebeard/commit/abd9908e01a8bbacd36cd71a98f02c19d0c14a39#diff-d6e7375760ed8e5a53c09150748c853f3b888c121db2dab161b10c36a9b6afef) but in a way that does not work on my end. 

I continued the migration to the best of my knowledge, and (optionally) added a benchmark for ancestors.

### DIFF
--- a/tests/test_benchmarks.py
+++ b/tests/test_benchmarks.py
@@ -23,13 +23,13 @@ def create_nodes(cls):
     """
     roots = []
     for i in range(5):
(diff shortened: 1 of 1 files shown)


solarissmoke:
Thanks!

Issue #418, opened 2026-09-01
`search_fields` and `list_filter` on a `TreeAdmin` only ever match root nodes


domibarton:
**Version:** 7.0.1 (also 7.0.0 — introduced with the admin rewrite)
**Django:** 6.0

## Description

On a `TreeAdmin` changelist, both the admin search box and any configured `list_filter` only ever consider **root nodes**. 
Anything at depth 2 or below is invisible to them, so searching for a node you can see by expanding the tree returns "0 results".

`search_fields` is effectively decorative on a `TreeAdmin` today, and a `list_filter` silently answers a different question than the one the user asked.

## Steps to reproduce

1. Register a model with `TreeAdmin` and give it `search_fields = ("desc",)`.
2. Create a root node `1` with a child `11`.
3. Open the changelist and search for `11`.

## Expected

The child node `11` is listed.

## Actual

No results. Only `1` is ever searchable.

The same applies to `list_filter`: filtering narrows the root nodes rather than the tree.

## Cause

`TreeAdmin.changelist_view()` sets `request._treebeard_parent_id = None` on every `GET`.
`TreeAdmin.get_queryset()` then returns `self.model.objects.get_root_nodes()`:

```python
def changelist_view(self, request, extra_context=None):
    if request.method == "GET":
        request._treebeard_parent_id = None
    return super().changelist_view(request, extra_context)

def get_queryset(self, request) -> QuerySet:
    ...
    if request._treebeard_parent_id:
        ...
    else:
        request._treebeard_parent = None
        qs = self.model.objects.get_root_nodes()
```

`ChangeList` applies the search and the filters on top of that, so they can only ever match the one level the changelist is showing. 
Deeper levels are reached through the AJAX `children/<parent_id>/` view, which the search never touches.

This is a consequence of the lazy-loading rewrite in 7.0.0.
Before it, the changelist queryset was the whole tree.

## Suggested fix

A searched or filtered changelist cannot be a drill-down view: there is no single level to drill into, and the matches deliberately come from all over the tree. 
So when the request carries a search term or a filter parameter, fall back to the full queryset and render a flat list:

- `TreeAdmin.get_queryset()` returns `super().get_queryset(request)` (and sets `request._treebeard_parent = None`) when the changelist is filtered.
- `tree_change_list.html` skips the `tree-context` `json_script` and the `treebeard-admin.js` include in that case, since neither expanding nor dragging means anything without the surrounding tree.
[...]

PR #419, opened 2026-09-01
Fix search & filter on the whole ModelAdmin tree

domibarton:
TreeAdmin.changelist_view() sets _treebeard_parent_id to None on every GET, so get_queryset() returns get_root_nodes(). ChangeList then applies the admin search and any list_filter on top of that, which means both can only ever match the level the changelist happens to be showing. Anything deeper is reached through the children/<parent_id>/ view, which neither of them touches, so search_fields was effectively decorative and list_filter silently answered a narrower question than it was asked.

A searched or filtered changelist cannot be a drill-down view: there is no single level to drill into, and the matches deliberately come from all over the tree. It now falls back to the full queryset and renders a flat list, dropping the tree context and its script, since neither expanding nor dragging means anything without the surrounding tree.

Paging and ordering do not count as filtering, so the tree survives both.

See the following issue for more information:
https://github.com/django-treebeard/django-treebeard/issues/418


treebeard/static/treebeard/treebeard-admin.js
```diff
@@ -217,8 +217,16 @@
     }
 
     $(document).ready(function () {
+        const contextElement = document.getElementById('tree-context');
+
+        // A filtered changelist is a flat list of matches with no tree to drag
+        // within, and renders no context for one.
+        if (contextElement === null) {
```
solarissmoke, line 224: Should this situation ever arise, given that the `script` tag is rendered conditionally in the template?

treebeard/templatetags/admin_tree.py
```diff
@@ -54,11 +54,13 @@ def tree_context(cl, request):
     as the `result_list` iterator.
     """
     qs = _get_annotated_context_queryset(cl)
-    depth = request._treebeard_parent.get_depth() + 1 if request._treebeard_parent else 1
+    # Absent when the changelist was never drilled into, e.g. on a filtered list.
+    parent = getattr(request, "_treebeard_parent", None)
```
solarissmoke, line 58: In what situation would the attribute be absent? We don't run this template tag at all if filtering is applied. I think the attribute is set in all code paths that would render this template, in `get_queryset()` in the changelist view.

solarissmoke (commented):
Hey @domibarton many thanks for the bug report and PR. A couple of small queries, otherwise looks great and I agree that dropping the tree is the right approach for filtered queries.

solarissmoke:
Merged manually in https://github.com/django-treebeard/django-treebeard/commit/b6b4fb9c29a1ec7a5b821d02ff3cf51de9ff833e with some tweaks. Thanks again!

